PR #80, opened 2015-01-07
Fix hyperlinks inside of table cells

h3xx:
This patch fixes a bug where a Hyperlink added to a Paragraph inside of
a Table Cell wasn't getting added to the Rels as well as getting
assigned a null id when being output as PowerPoint2007.

What this patch does is when adding Hyperlinks to the Rels, it also
iterates over Paragraphs inside of Table Cells as if they were inside of
a normal RichText shape.

### DIFF
--- a/src/PhpPowerpoint/Writer/PowerPoint2007/Rels.php
+++ b/src/PhpPowerpoint/Writer/PowerPoint2007/Rels.php
@@ -327,8 +327,24 @@ class Rels extends AbstractPart
                 }
 
                 // Hyperlink on rich text run
-                if ($iterator->current() instanceof RichText) {
-                    foreach ($iterator->current()->getParagraphs() as $paragraph) {
+                if ($iterator->current() instanceof PHPPowerPoint_Shape_RichText ||
+                    $iterator->current() instanceof PHPPowerPoint_Shape_Table
+                ) {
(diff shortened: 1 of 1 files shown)


Progi1984:
@h3xx Could you look at the develop branch of PHPPowerPoint ? It seems [that bug has been fixed](https://github.com/PHPOffice/PHPPowerPoint/issues/70#issuecomment-67153785).

Progi1984:
@h3xx Some news ?

h3xx:
@Progi1984 I'm on an older version at my workplace. However the branch doesn't exist in github. Could you please create the branch for version 0.1.0 so I can merge this fix into that?

As it is, this is still a relevant fix for that version.

Progi1984:
@h3xx There is no branch/tag for 0.1.0 version. I advice you to update your version to recent for having another bugfixes : https://github.com/PHPOffice/PHPPowerPoint/releases.

h3xx:
But 0.1.0 is the current release according to the [download page](http://phppowerpoint.codeplex.com/releases/view/26621).

If this is incorrect, is it possible to take it down to avoid confusion?

Progi1984:
Since the 01/01/2014, PHPOffice is not anymore on CodePlex and have migrated to Github.

Reference : http://phppowerpoint.codeplex.com/

PR #94, opened 2015-03-08
Fix slide layout out of order

kenliau:
This PR fixes the issue where wrong slide layout was being used to generate powerpoint. 
The root cause was that, in slideMaster1.xml.rels, it was assuming all the slideLayouts are in the same order as the actual slides in a powerpoint. But in reality, the slideLayouts were not in sorted order causing the generation to use a different layout. 

I have narrow down to 2 different solutions and submitting two different PR as there are pros and cons in both.

This PR is Solution#1.

###### Solution
1. KSort AbstractLayoutPack $layout so we can still findLayoutIndex and use ($layoutIndex + 1) in Rels.php to generate slides.

	* Pro: 1 line of change.
	* Con: Ksorting the list of slide layouts might affect performance if we have large number of slide layouts.

2. Do not sort $layout, but refactor findLayoutIndex to findLayoutId in AbstractLayoutPack.php. Then we can replace findLayoutIndex by findLayoutId and use ($layoutId) instead of ($layoutIndex + 1) in Rels.php to generate slides.
	
	* Pro: Improves readability as we use layoutId instead of layoutIndex.
	* Con: Couple line of changes in 2 different files.

Again, this PR is Solution#1. I have also added unit tests for my changes. Unfortunately, I am not able to completely implement testConstruct for TemplateBasedTest.php

Solution#2 PR is [here](https://github.com/PHPOffice/PHPPowerPoint/pull/95).
Please critique and provide feedback as this is my very first PR to the open source community.


Progi1984:
I prefer the #95, which brings more flexibility.

PR #193, opened 2016-03-08
#168 : Support for images in base64

Progi1984:


sergiodebcn:
Thanks, I am going to use this feature.

PR #178, opened 2016-01-28
Add appear animation to phppresentation for Powerpoint2007 writer

JewrassicPark:
Added animation class, added animation tags to Powerpoint2007 Slide Writer, added animation objects to Slide.php

This only supports appear. Shapes can be grouped to appear together or sequentially by order they're added as animations. 

Use:

```
$animation = new Animation(); //create the animation
$currentSlide = $this->pp->getActiveSlide();  //get the current slide
$currentSlide->addAnimation($animation); //add the animation to the current slide
/*
* add n shapes to animation
* every shape in an animation will appear at the same time on click
*/
$animation->addShape($shape); 
$animation->addShape($shape2); 
```

PR #214, opened 2016-03-31
Adding animations

JewrassicPark:
Updated for latest dev branch

Previous : #178

### DIFF
--- a/src/PhpPresentation/Writer/PowerPoint2007/PptSlides.php
+++ b/src/PhpPresentation/Writer/PowerPoint2007/PptSlides.php
@@ -476,7 +476,7 @@ class PptSlides extends AbstractDecoratorWriter
         $objWriter->endElement();
 
         // Loop shapes
-        $shapeId = 0;
+        $shapeId = 1;
         $shapes  = $pSlide->getShapeCollection();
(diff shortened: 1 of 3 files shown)


JewrassicPark:
I'm not really sure why I had to do this again with the latest dev branch but here it is.

src/PhpPresentation/Writer/PowerPoint2007/PptSlides.php
```diff
@@ -476,7 +476,7 @@ public function writeSlide(Slide $pSlide)
         $objWriter->endElement();
 
         // Loop shapes
-        $shapeId = 0;
+        $shapeId = 1;
```
Progi1984: @JewrassicPark Why do you modify this value ?

JewrassicPark: I answered this in the last fork.

> It seemed like powerpoint was treating shape 0 as the slide itself. IIRC I couldn't get the animations to work when the shape index started at 0.

Progi1984: Oops, I forgot it :)

Issue #251, opened 2016-08-18
README.md missing requirements


danielbair:
The README.md does not mention the PHPOffice/Common as a requirement for PHPPresentation.
Thus if one is downloading PHPPresentation source and attempting to use the samples it will throw errors about PHPOffice/Common/... file not found.

danielbair:
I have created a pull request here... https://github.com/PHPOffice/PHPPresentation/pull/254

Progi1984:
Fixed in the commit dbc34df.

Progi1984:
And documentation added : http://phppresentation.readthedocs.io/en/develop/installing.html

PR #259, opened 2016-09-08
#251 : Documentation for manual install

Progi1984:


Harmonitron:
https://phppowerpoint.readthedocs.io/en/stable/installing.html, manual install, gives instructions for download and two 'require_once' for the Autoloader.php. It doesn't say where these two lines are to be installed. Can you clarify?
Many thanks.

Issue #268, opened 2016-09-15
bar charts are not honouring the setLabelPosition


pgee70:
Hi, this one is easy!
the PhpOffice\PhpPresentation\Writer\PowerPoint2007\PptCharts.php::writeTypeBar does not honour series:setLabelPosition.

in line 982 adding:

``` php
$this->writeElementWithValAttribute($objWriter, 'c:dLblPos', $series->getLabelPosition());
```

seems to fix the issue (i just looked at the code for writeTypePie to see this.
##

Progi1984:
@pgee70 Hi, Could you give me a screenshot with the bug ? Thanks.

pgee70:
Thanks for responding.
It is not something taking a screenshot of can help. The code isn't implemented so the feature doesn't work.
The column label position just goes to the default position - I wanted it to go in the end of the label and it wouldn't work.

When that line is added the label position behaves as expected.

Progi1984:
@pgee70 I just modified your fix. Could you test the PR #289 ?

Progi1984:
@pgee70 Some feedback, please.

Progi1984:
@pgee70 Some feedback, please.

Progi1984:
@pgee70 Fixed in develop branch

PR #289, opened 2016-10-27
#268 : Fixed the label position in bar chart

Progi1984:
CC: #268


pgee70:
Hi I really would like to help but don't understand github well enough to do it. If the setLabelPosition is being honoured, and the labels can be printed inside the top of the column charts then call it fixed!

PR #307, opened 2017-01-08
PowerPoint2007 writer schema validation test case

k42b3:
Hi, this adds a new schema test case which validates the produced XML from the `PowerPoint2007`
writer against the ECMA/ISO `Office Open XML File Formats` XSD. In the test (tests/PhpPresentation/Tests/Writer/PowerPoint2007/SchemaTest.php) we basically generate a ppt
based on the sample files and then check whether the generated XML is valid against the schema. This should help to protect us from generating the wrong format. Of course this test covers only the elements which are used but it is a start.

The test brought up an issue: Microsoft generates percentage values formatted as 1000th of a percent
but the standard requires it to be formatted with a trailing percent sign. I have adjusted the writer so that
the values are generated accordingly to the schema.
See: https://msdn.microsoft.com/en-us/library/gg548608(v=office.12).aspx


Progi1984:
@k42b3 Could you update the Changelog.md for adding your improvements, please ? Have you another modifications in this PR ?

k42b3:
Hi, ok I have updated the changelog. So basically it adds only the test case and changes percentage format.

Progi1984:
@k42b3 Very good job. I just merged it.

k42b3:
Thats great, thx!

Issue #273, opened 2016-09-20
There are no options to change margins in cells


alejandrososa:
Good morning, I searched all the documents which indicate how to modify the margins of a cell in a table, but I have not found anything. It would be interesting to contemplate the library to change the margins to a cell, because in my case I'm building the table has many columns and would help to take the edge that comes by default.

Thanks for the work they have done.
![margenes](https://cloud.githubusercontent.com/assets/13384058/18663115/7973b3ee-7f1d-11e6-9dd8-11681457a9b1.png)
##

ellisgl:
Yeah - there's a Right / Left margin for paragraphs, but not for cells.

Progi1984:
Fixed in develop branch ✔️ (w/ unit tests and [documentation](http://phppresentation.readthedocs.io/en/develop/shapes_table.html#define-margins-of-a-cell))

PR #315, opened 2017-01-24
Allow setting top and bottom margins, write table margins in PPTX writer

carlosafonso:
This PR allows setting bottom and top margins, and propagates those margins (as well as left and right) when serializing table objects.

I think this would cover what is requested in #273.


ellisgl:
This would be nice to have!

Progi1984:
Fixed in develop branch ✔️ (w/ unit tests and [documentation](http://phppresentation.readthedocs.io/en/develop/shapes_table.html#define-margins-of-a-cell))

ajmorenodelarosa:
Niceeee,  awesome feature!

PR #322, opened 2017-02-10
fix arraystyles being empty

tcitworld:
Sometimes 
```php
$this->arrayStyles[$keyStyle]['listStyle']
```
is empty but 
```php
$this->arrayStyles[$keyStyle]
```
is set. So it leads to an undefined offset error.


tcitworld:
cc @Progi1984

Progi1984:
Thanks for the MR @tcitworld

PR #360, opened 2017-05-17
Fix. Text is subscripted when set superscript to false

qmachard:
Hi,

I fixed a little bug : in a Font, when I set the subscript to false, the superscript is automatically set to true. And, when I set the superscript is the opposite. 

I also modified the test case to test this toggle for subscript and superscript.

Good Job!
Bye


whldk:
isSubScript and isSuperScript  Always false when getting

Issue #382, opened 2017-07-11
Charts are broken / doesn't validate against Office 2013 XML


simplyray:
I'm using the latest 0.9.0 release but the file generated from the Sample_05_Chart.php is broken and Powerpoint 2016 needs to repair the file before opening it (which removes the charts completly). 

[Sample_05_Chart.pptx](https://github.com/PHPOffice/PHPPresentation/files/1138865/Sample_05_Chart.pptx)

I've tried to validate the generated *.pptx file against Office 2013 XML formats and it throws 218 errors. Seems like every value for the attribute 'val' is a _string_ having a percentage sign behind it instead of _Int32_

Progi1984:
@simplyray Thanks for the issue.

Where can i find the xsd for validating pptx file against Office 2013 XML formats ? (And may be Office 2007 & 2010)

simplyray:
@Progi1984 I've used the [Open XML SDK Tool](https://www.microsoft.com/en-us/download/details.aspx?id=30425) from Microsoft

Progi1984:
Todo :
* Update [tests schema ooxml](https://github.com/PHPOffice/PHPPresentation/tree/develop/tests/resources/schema/ooxml) from [libreoffice/contrib/test-files](https://cgit.freedesktop.org/libreoffice/contrib/test-files/tree/ooxml-strict)
* Test all files with ooxml if Presentation2007 as writer

Progi1984:
PR relative to this issue : #390

larryb-redflare:
See comment on #353 
PhpPresentation\Writer\PowerPoint2007\PptCharts Line 102 outputs a % which cause invalid XML

  $objWriter->writeElementIf($hPercent != null, 'c:hPercent', 'val', $hPercent . '%');

Progi1984:
@dlollman You can test this branch and give your feedback.

dlollman:
@Progi1984 Thanks, no good for that branch for me. Using Powerpoint for mac (microsoft). Relevant code I am using to test this out:

`$current_slide++;
    // Slide 5 (REDACTED)
    // Create slide
    ////echo date('H:i:s') . ' Create slide '.$current_slide.EOL;
    $slides[$current_slide] = $objPHPPresentation->createSlide();
    $slides[$current_slide]->setName('REDACTED');
    $slides[$current_slide]->setBackground($oBkgImageSlide);

    //Title of slide
    $richtext = $slides[$current_slide]->createRichTextShape()
[...]

PR #384, opened 2017-07-12
Fix to match Office 2013 XML

simplyray:
First part of the fix for issue #382, PptCharts.php still needs some major adjustments (nesting from ChartShapeProperties (c:spPr) having fill or outline breaks it)


simplyray:
Just double-checked: removing the % in writeColor leads to issues with TextRuns beeing transparent.

Progi1984:
@simplyray Finally no modification against develop branch (since #382 / #390)

simplyray:
> @simplyray Finally no modification against develop branch (since #382 / #390)

Thank you very much! I'll update as soon as possible!

PR #417, opened 2017-11-04
Use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase

carusogabriel:
I use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCases. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

Just needed to bump PHPUnit version to [`4.8.36`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-4.8.md#4836---2017-06-21) for compatibility with `namespaces`.


Progi1984:
@carusogabriel Thank you for this huge PR.

PR #433, opened 2017-12-29
Refactoring tests

carusogabriel:
I've refactored some tests using `PHPUnit` assertion methods, instead of `PHP`'s. This will give us better error messages in case of failure.


Progi1984:
@carusogabriel Good job again. Thank you !

PR #453, opened 2018-02-08
Remove % signs from PptCharts to fix chart generation. Fixes #452

### LINKED ISSUE #452: Percent signs in PptCharts cause unrepairable failure
When using PowerPoint2007 writer to create charts and opening in Powerpoint for Mac, I get an unrepairable error from PPT. 

The source of the issue is two lines in PptCharts: 

Line 108: 

```
$objWriter->writeElementIf($hPercent != null, 'c:hPercent', 'val', $hPercent . '%');
```

Line 2325:

```
$objWriter->writeAttribute('val', '100%');
```

If I remove the two percent symbols everything works as expected.

Would this be a safe pull request to make?

bkanber:
I have removed two instances of % signs in PptCharts which cause presentations with charts to be corrupted on newer versions of Powerpoint. I have also updated the test suite to match. 

However I am unsure if the changes I've made to the test schema are acceptable. Please let me know and I can updated.

Fixes issue #452 and possibly a couple others.

### DIFF
--- a/src/PhpPresentation/Writer/PowerPoint2007/PptCharts.php
+++ b/src/PhpPresentation/Writer/PowerPoint2007/PptCharts.php
@@ -105,7 +105,7 @@ class PptCharts extends AbstractDecoratorWriter
 
         // c:hPercent
         $hPercent = $chart->getView3D()->getHeightPercent();
-        $objWriter->writeElementIf($hPercent != null, 'c:hPercent', 'val', $hPercent . '%');
+        $objWriter->writeElementIf($hPercent != null, 'c:hPercent', 'val', $hPercent);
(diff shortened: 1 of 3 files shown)


bkanber:
Regarding the changes I made to the test schema: I looked up the schema here http://www.datypic.com/sc/ooxml/s-dml-chart.xsd.html and added the xsd:unsignedShort to the union statement, as both of the types in question are listed as being derived from xsd:unsignedShort. I have left in the percent sign types in the union, figuring that this would allow for some backwards compatibility in validation, however let me know if you would like me to update that schema. I'm just not really sure what it's properly supposed to be.

PR #467, opened 2018-04-13
Chart referenceType guessing update, Set tickLabels to 'low' by defau…

samverhoeven:
…lt, fix x axis for scatterCharts, show correct number format on y and x axis


Progi1984:
@samverhoeven Hi why do you close this PR ?

PR #555, opened 2019-08-27
Defining width & height of a shape don't return any error if width & height were equal to 0

surger:
Add zero check before divizion

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bugfix
 - PowerPoint2007 : Text is subscripted when set superscript to false - @qmachard GH-360
+- Core : Defining width & height of a shape don't return any error if width & height were equal to 0 - @surger GH-555
 
 ### Changes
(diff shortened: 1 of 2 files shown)


Progi1984:
Hi @surger, thank you for your contribution. Could you update the CHANGELOG.md and add a test in tests/PhpPresentation/Tests/Shape/AbstractGraphicTest.php before merge ? Thanks

surger:
@Progi1984 Which version should I specify CHANGELOG.md? And my commit so little, you may miss my commit :).
And I don't think what my changes need test.
AbstractGraphicTest.php have enough tests for it.

Progi1984:
@surger The next version is 0.10. No, each commit, each contribution is a part of next version.

Ok for no tests. You're right.

Progi1984:
@surger Could you apply the same logic to the method setWidth ? And update Changelog.md 😉 After that I will merge it.

Progi1984:
Thank you @surger for your contribution.

Issue #570, opened 2020-03-02
Extracted timestamps in document properties are incorrect


czarpino:
### What happened
Extracting document properties `created` and `modified` both return the current date.

### Expected
Values should reflect what is in the respective tags of created and modified properties in docProps/core.xml.

### Steps to reproduce

1. Prepare some pptx file
2. Copy file into a .zip file e.g. `cp some.pptx some.zip`
3. Inspect zip
    * Go inside docProps > core and check dcterms:created & dcterms:modified
    * Maybe use some tool to convert the displayed date into a timestamp e.g. https://www.epochconverter.com
4. Extract the created and modified properties using the library
    ```php
    $pptReader = new PowerPoint2007();
    $phpPpt = $pptReader->load('some.pptx');
    $docInfo = $phpPpt->getDocumentProperties();

    echo sprintf('Created timestamp: %d', $docInfo->getCreated());
    echo sprintf('Created timestamp: %d', $docInfo->getModified());
    ```
5. Extracted timestamp from library should not be the same as the values derived from core.xml.

Progi1984:
Fixed on `develop` branch. :) Thanks for the report.

PR #571, opened 2020-03-02
#570 Use new DateTime instance returned

czarpino:
DateTime::createFromFormat does not modify the current DateTime instance
used.

Addresses: https://github.com/PHPOffice/PHPPresentation/issues/570


Progi1984:
Already fixed on `develop` branch. Thanks for the PR.

PR #609, opened 2020-12-20
Change samples to run without composer installation (with your own installation)

pal-software:
1. Now samples run only if PHPPresentation was installed with composer. So its look if vendor directory is exists.
Change the samples to run when PHPPresentation was installed without composer (with your own installation).

Then was downloaded PHPOffice/Common release from its releases page and unpack to PHPOffice directory. So directory structure must be:
<pre>
PHPOffice
    Common   <-- PHPOffice/Common release
        src
            Common
    PHPPresentation    <-- PHPOffice/PHPPresentation release
        samples
        src
</pre>

2. Added results folder to samples folder such as samples don't running without results folder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,5 +24,4 @@ vendor
 /samples/Sample_00_Test.php
 /samples/#*
 /samples/Github_*.*
-/samples/results
 /composer.lock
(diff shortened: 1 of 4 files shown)


Progi1984:
Superseeded by #784

PR #614, opened 2021-01-11
Fix #613: Replace PHPExcel with PhpSpreadsheet

### LINKED ISSUE #613: setIncludeSpreadsheet => PHPExcel has not been loaded
If I include the data of a chart:

    $shape->setIncludeSpreadsheet(true);

I get the error:

    PHPExcel has not been loaded. Include PHPExcel.php in your script, e.g. require_once 'PHPExcel.php'. 

This is caused by [PptCharts::writeSpreadsheet](https://github.com/PHPOffice/PHPPresentation/blob/1fcc84072ee0061f8991b8f83050a99e706fe5ce/src/PhpPresentation/Writer/PowerPoint2007/PptCharts.php#L213)

It tries to use PHPExcel that is now PhpSpreadsheet. Any hint how to fix this?

kw-pr:
setIncludeSpreadsheet() still uses PHPExcel instead of PhpSpreadsheet.
See also issue #613


Progi1984:
Thanks @kw-pr.

Done in #652

Issue #634, opened 2021-05-31
Roadmap for new version


Progi1984:
- [x] Migration to Github Actions from Travis CI
    - #635 
- [x] Support for recents versions of PHP and remove support for PHP < 7.1
    - #636
- [x] Add PHPStan
    - #639
- [x] Add PHPCSFixer
    - #637
    - #640 
- [x] Use Coveralls for replacing Scrutinizer
    - #648
- [x] Remove all deprecated
    - #650
- [x] Migration documentation to Markdown (@mkdocs)
    - #647
    - #649
- [x] Migrate from PHPExcel to PhpSpreadsheet
    - #652
- [x] Define new exceptions for each error
    - #677 
- [x] Add declare strict_types for each file
    - #673
- [x] Upgrade major version of PHPOffice\Common
    - #679

PR #637, opened 2021-06-06
#634 : Enable PHPCSFixer 

Progi1984:


Brenneisen:
@Progi1984 Due to the change I get an error with SlideLayouts:

`PhpOffice\PhpPresentation\Slide\AbstractSlide::getRelsIndex(): Return value must be of type string, null returned.`

It seems that `relsIndex` is not set for SlideLayouts and remains null. By adjusting the return type it does not work now. The method is used in:

`PhpOffice\PhpPresentation\Writer\PowerPoint2007\PptSlideLayouts::67`

Brenneisen:
@Progi1984 I have created a PR for it: https://github.com/PHPOffice/PHPPresentation/pull/646

Brenneisen:
@Progi1984 Could you please merge this PR?

PR #664, opened 2021-08-02
Fix column indices for embedded spreadsheets

michael-roth:
This fixes a bug with embedded spreadsheet data.

**Reproduction**
1. Use code below to reproduce
2. Open generated presentation file
3. Edit chart data to open embedded spreadsheet

```php
use PhpOffice\PhpPresentation\IOFactory;
use PhpOffice\PhpPresentation\PhpPresentation;
use PhpOffice\PhpPresentation\Shape\Chart\Series;
use PhpOffice\PhpPresentation\Shape\Chart\Type\Bar;

$objPHPPowerPoint = new PhpPresentation();
$currentSlide = $objPHPPowerPoint->getActiveSlide();

$chart = new Bar();
$series = new Series('Series 1', [
    'A' => 90,
    'B' => 80,
    'C' => 50,
    'D' => 67,
    'E' => 99,
    'F' => 23,
]);
$chart->addSeries($series);

$chartShape = $currentSlide->createChartShape();
$chartShape->setIncludeSpreadsheet(true);
$chartShape->getPlotArea()->setType($chart);
$chartShape
    ->setWidth(400)
    ->setHeight(300)
    ->setOffsetX(100)
    ->setOffsetY(60)
;

$oWriterPPTX = IOFactory::createWriter($objPHPPowerPoint, 'PowerPoint2007');
$oWriterPPTX->save(__DIR__ . "/sample.pptx");
```

**Expected result**
Chart data is editable, chart won't break

**Actual result**
Chart data is not editable or chart labels break

**Cause**
The bug is caused by wrong column indices in the chart writer. PhpSpreadsheet uses one-based indices for columns while zero-based indices are used in the writer.

### DIFF
--- a/src/PhpPresentation/Shape/Chart/Type/Area.php
+++ b/src/PhpPresentation/Shape/Chart/Type/Area.php
@@ -39,9 +39,4 @@ class Area extends AbstractType implements ComparableInterface
 
         return md5($hash . __CLASS__);
     }
-
-    public function getXmlElementName(): string
-    {
-        return 'c:areaChart';
-    }
 }
--- a/src/PhpPresentation/Shape/Chart/Type/Area.php
+++ b/src/PhpPresentation/Shape/Chart/Type/Area.php
@@ -39,4 +39,9 @@ class Area extends AbstractType implements ComparableInterface
 
         return md5($hash . __CLASS__);
     }
+
+    public function getXmlElementName(): string
+    {
+        return 'c:areaChart';
+    }
 }
--- a/tests/PhpPresentation/Tests/Writer/PowerPoint2007/PptChartsTest.php
+++ b/tests/PhpPresentation/Tests/Writer/PowerPoint2007/PptChartsTest.php
@@ -107,6 +107,7 @@ class PptChartsTest extends PhpPresentationTestCase
         $this->assertZipXmlElementExists('ppt/charts/' . $oShape->getIndexedFilename(), $element);
         $element = '/c:chartSpace/c:chart/c:plotArea/c:lineChart/c:ser';
         $this->assertZipXmlElementExists('ppt/charts/' . $oShape->getIndexedFilename(), $element);
+        $this->assertZipXmlElementCount('ppt/charts/' . $oShape->getIndexedFilename(), $element, 1);
         $element = '/c:chartSpace/c:chart/c:plotArea/c:lineChart/c:ser/c:tx/c:v';
         $this->assertZipXmlElementNotExists('ppt/charts/' . $oShape->getIndexedFilename(), $element);
         $element = '/c:chartSpace/c:chart/c:plotArea/c:lineChart/c:ser/c:tx/c:strRef';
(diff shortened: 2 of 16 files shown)


Progi1984:
@michael-roth Great fix. Could you rebase your PR to the `develop` branch and apply to `RadarChart` ?

I really need to add more unit tests for testing this. Could you fix you can add it ?

michael-roth:
@Progi1984 Thanks. I rebased, fixed the remaining lines and added tests to check the series values. 

What do you think? Let me know if you need more.

Progi1984:
@michael-roth Yeap, I have a last thing to test and it's ok. I add comment in code. :heart_eyes: on your bugfix.

tests/PhpPresentation/Tests/_includes/PhpPresentationTestCase.php
```diff
@@ -247,6 +247,28 @@ public function assertZipXmlElementExists($filePath, $xPath): void
         );
     }
 
+    /**
+     * @param string $filePath
+     * @param string $xPath
+     * @param int $index
+     */
+    public function assertZipXmlElementAtIndexExists($filePath, $xPath, $index): void
```
Progi1984, line 255:
```suggestion
    public function assertZipXmlElementAtIndexExists(string $filePath, string $xPath, int $index): void
```

tests/PhpPresentation/Tests/Writer/PowerPoint2007/PptChartsTest.php
```diff
@@ -107,20 +107,26 @@ public function testChartIncludeSpreadsheet(): void
         $this->assertZipXmlElementExists('ppt/charts/' . $oShape->getIndexedFilename(), $element);
         $element = '/c:chartSpace/c:chart/c:plotArea/c:lineChart/c:ser';
         $this->assertZipXmlElementExists('ppt/charts/' . $oShape->getIndexedFilename(), $element);
+        $this->assertZipXmlElementCount('ppt/charts/' . $oShape->getIndexedFilename(), $element, 1);
```
Progi1984, line 110: Could you for `testChartIncludeSpreadsheet` a `dataProvider` which pass in parameter 
- each chartType (Line/Radar....) where you have made a change in `PptCharts`
- a string for element xml `c:lineChart` relative to each chartType
?

That allows me to check that each change in `PptCharts` is ok.

Progi1984 (commented):
A last feedback

michael-roth:
@Progi1984 I worked on your feedback and changed a few things:

1. Added abstract method `getXmlElementName` to `AbstractChartType` and implemented it in every chart type. This way I did not have to use it in the data provider and we don't have repetition of the constant strings.

2. The order of elements was wrong for Doughnut and Radar charts, because of the defined `<xsd:sequence>`. I moved these elements in the writer methods to fix it.
For Radar charts I removed the `c:smooth` element. Radar charts seem to not support the smooth config.

3. For Scatter charts there is a different xml path for the value cache (`c:yVal` instead of `c:val`), so I added a check in the unit test.

src/PhpPresentation/Shape/Chart/Type/Area.php
```diff
@@ -39,4 +39,9 @@ public function getHashCode(): string
 
         return md5($hash . __CLASS__);
     }
+
+    public function getXmlElementName(): string
```
Progi1984, line 43: Could you remove `getXmlElementName` method for all chart types ? It's not compatilble with `Open Documentation Format` for `LibreOffice`.

michael-roth: Sorry, I reverted my change in the chart types and added the xml element names to the data provider.

Let me know if you need more.

mindline-analytics:
Hi @Progi1984,
Hi @michael-roth,

I am waiting for this PR to be merged, as I use the embedded spreadsheets a lot.
Is there anything I can help with to get it ready to be merged? Unfortunately, I have no idea why phpstan fails on the dataProvider.

Progi1984:
Superseeded by #773

PR #671, opened 2021-08-20
Fix issue when first element in series is null

ksmeeks0001:
Creating charts fail when the first element of a series is null. This fix will check the data type of a series based on the first non-null element of the list rather than the first element.

### DIFF
--- a/src/PhpPresentation/Writer/PowerPoint2007/PptCharts.php
+++ b/src/PhpPresentation/Writer/PowerPoint2007/PptCharts.php
@@ -338,7 +338,16 @@ class PptCharts extends AbstractDecoratorWriter
         // c:strLit / c:numLit
         // c:strRef / c:numRef
         $referenceType = ($isReference ? 'Ref' : 'Lit');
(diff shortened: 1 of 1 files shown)


src/PhpPresentation/Writer/PowerPoint2007/PptCharts.php
```diff
@@ -338,7 +338,16 @@ protected function writeMultipleValuesOrReference(XMLWriter $objWriter, bool $is
         // c:strLit / c:numLit
         // c:strRef / c:numRef
         $referenceType = ($isReference ? 'Ref' : 'Lit');
-        $dataType = is_numeric($values[0]) ? 'num' : 'str';
+        
+        // get data type from first non-null value
+        $dataType = 'num';
+        foreach($values as $value){
+            if ( !isset($value) ){
```
Progi1984, line 345:
```suggestion
            if (!isset($value)) {
```

Progi1984:
@ksmeeks0001 Could you allow me to contribute to your repository (I have brought some improvements & tests to your PR) ?

Progi1984:
Superseeded by #778

PR #684, opened 2021-11-19
Propose some contributions developed for my project

devX2712:
My project use a Powerpoint file as template and duplicate a slide to generate a final file with multiple slides including some specific datas.
I'm very happy to contribute to this project.
It was very useful to find this library and the to do use it.
I hop you will includes them into the master branch.
Because I use some shape to do some specific action, the name is necessary. So I add the search function and extend name attribute to all shape.


Progi1984:
Superseeded by #788.

PR #696, opened 2022-01-27
fix for loadMasterSlide

berbeflo:
in the method loadMasterSlide was a call to Font::setSize which expects an integer
while the call passed a float which resulted in an TypeError


Progi1984:
Fixed in #766. Thank you for your contribution.

Issue #681, opened 2021-10-22
Video embed is broken


crazysix:
The video embed appears to be broken. On the dev branch, if I run Sample_03_Video.php and attempt to open the pptx I get an error.
<img width="264" alt="Screen Shot 2021-10-21 at 7 43 52 PM" src="https://user-images.githubusercontent.com/1428324/138379803-7ebbcd0c-e581-4eaf-8e1c-f9f0f2133f85.png">

The xml looks a bit different than a slide that I manually add a video to. The odp file is also not working, but I am mostly looking at pptx. I am on PowerPoint 16.54 on a mac. I am attaching my sample (testvideo.pptx) and the output from the sample (Sample_03_Video.pptx).
[Sample_03_Video.pptx](https://github.com/PHPOffice/PHPPresentation/files/7393770/Sample_03_Video.pptx)
[testvideo.pptx](https://github.com/PHPOffice/PHPPresentation/files/7393772/testvideo.pptx)

potofcoffee:
I can confirm this is broken. 

At least part of the problem is here:
https://github.com/PHPOffice/PHPPresentation/blob/f5907e199c6b4e52e574a9febcfd212e869ffbca/src/PhpPresentation/Writer/PowerPoint2007/AbstractSlide.php#L1318

This won't create a correct relationship id, since ````$shape->relationId```` is a string like ````rId2```` and can't be cast into an int.

However, that's only part of the problem. When I edit this manually to the correct relationship id, PowerPoint will open the presentation, but say that the video file was not found.

potofcoffee:
Correction: With the above fix, PowerPoint finds and correctly plays the video. The only thing not working is a thumbnail display when the video is stopped. This would require another relationship with a thumbnail image. However, that's a big step which would probably introduce new dependencies to automatically create thumbnails from videos when packing the .pptx file.

zmip:
Did this ever work? And were thumbnails created automatically back when it worked?

PR #697, opened 2022-01-31
Fix broken video file relationship

potofcoffee:
This fixes the broken relationship reference for video files as described in #681. Videos will work with this patch, but no thumbnail is displayed when the videos is stopped. (This would require thumbnail creation and another relationship for the thumbnail image.)


potofcoffee:
Hello @Progi1984, is this repository still alive? It would be a real shame, if not, after all the great work you have done here.
Is there anything I can do to help this pull request go through?

potofcoffee:
@Progi1984 Now everything should be as expected.

Issue #354, opened 2017-04-20
Readers do not handle invalid images


MisatoTremor:
When the readers create ``Drawing\Gd`` and use imagecreatefromstring with an unsupported or else invalid format it returns false. They then just pass this to ``setImageResource`` without verification.
When ``getContents`` on these object is called (i.e. in a writer) it obviously can't return a valid image.
Should this be fixed and if so, what would be the best way? Just don't create the shape at all?

PR #707, opened 2022-04-11
ensure that imagecreatefromstring returns valid image before calling setImageResource

aelliott1485:
fixes Issue #354


Progi1984:
Superseeded by #775

PR #713, opened 2022-07-24
fix IncludedSpreadsheet column num error

zhengwhizz:
IncludedSpreadsheet data was wrong columns, so when edit the embed data, the cat name will wrong.


Progi1984:
Superseeded by #774

Issue #751, opened 2023-08-14
Notes in pptx gives error on reader


Offbeatmammal:
if I have an Office365 edited PPTX with no notes, I am able to open and enumerate the slides with no issue.
if however I have notes on the slides I get the error below.

Able to repro using the sample_12.pptx from here (before works fine, add a simple letter to the Notes on slide1 and it breaks. removing the Note doesn't help).
````
require_once 'vendor/autoload.php';
use PhpOffice\PhpPresentation\IOFactory;

$oReader = \PhpOffice\PhpPresentation\IOFactory::createReader('PowerPoint2007');
$oPresentation04 = $oReader->load('Sample_12.pptx');
````

Fatal error: Uncaught TypeError: PhpOffice\PhpPresentation\Reader\PowerPoint2007::loadShapeRichText(): Argument #3 ($oSlide) must be of type PhpOffice\PhpPresentation\Slide\AbstractSlide, PhpOffice\PhpPresentation\Slide\Note given, called in C:\AP-Web\engine-word\vendor\phpoffice\phppresentation\src\PhpPresentation\Reader\PowerPoint2007.php on line 1375 and defined in C:\AP-Web\engine-word\vendor\phpoffice\phppresentation\src\PhpPresentation\Reader\PowerPoint2007.php:888
Stack trace:
#0 C:\AP-Web\engine-word\vendor\phpoffice\phppresentation\src\PhpPresentation\Reader\PowerPoint2007.php(1375): PhpOffice\PhpPresentation\Reader\PowerPoint2007->loadShapeRichText(Object(PhpOffice\Common\XMLReader), Object(DOMElement), Object(PhpOffice\PhpPresentation\Slide\Note))
#1 C:\AP-Web\engine-word\vendor\phpoffice\phppresentation\src\PhpPresentation\Reader\PowerPoint2007.php(755): PhpOffice\PhpPresentation\Reader\PowerPoint2007->loadSlideShapes(Object(PhpOffice\PhpPresentation\Slide\Note), Object(DOMNodeList), Object(PhpOffice\Common\XMLReader))
#2 C:\AP-Web\engine-word\vendor\phpoffice\phppresentation\src\PhpPresentation\Reader\PowerPoint2007.php(369): PhpOffice\PhpPresentation\Reader\PowerPoint2007->loadSlideNote('notesSlide1.xml', Object(PhpOffice\PhpPresentation\Slide))
[...]

Issue #710, opened 2022-05-06
PowerPoint2007::loadShapeRichText() must be an instance of AbstractSlide, instance of Note given


Sarke:
```
TypeError: Argument 3 passed to PhpOffice\PhpPresentation\Reader\PowerPoint2007::loadShapeRichText() must be an instance of PhpOffice\PhpPresentation\Slide\AbstractSlide, instance of PhpOffice\PhpPresentation\Slide\Note given, called in /app/vendor/phpoffice/phppresentation/src/PhpPresentation/Reader/PowerPoint2007.php on line 1387 and defined in /app/vendor/phpoffice/phppresentation/src/PhpPresentation/Reader/PowerPoint2007.php:886

Stack trace:
#0 /app/vendor/phpoffice/phppresentation/src/PhpPresentation/Reader/PowerPoint2007.php(1387): PhpOffice\PhpPresentation\Reader\PowerPoint2007->loadShapeRichText(Object(PhpOffice\Common\XMLReader), Object(DOMElement), Object(PhpOffice\PhpPresentation\Slide\Note))
#1 /app/vendor/phpoffice/phppresentation/src/PhpPresentation/Reader/PowerPoint2007.php(756): PhpOffice\PhpPresentation\Reader\PowerPoint2007->loadSlideShapes(Object(PhpOffice\PhpPresentation\Slide\Note), Object(DOMNodeList), Object(PhpOffice\Common\XMLReader))
#2 /app/vendor/phpoffice/phppresentation/src/PhpPresentation/Reader/PowerPoint2007.php(370): PhpOffice\PhpPresentation\Reader\PowerPoint2007->loadSlideNote('notesSlide1.xml', Object(PhpOffice\PhpPresentation\Slide))
#3 /app/vendor/phpoffice/phppresentation/src/PhpPresentation/Reader/PowerPoint2007.php(170): PhpOffice\PhpPresentation\Reader\PowerPoint2007->loadSlides('<?xml version="...')
#4 /app/vendor/phpoffice/phppresentation/src/PhpPresentation/Reader/PowerPoint2007.php(137): PhpOffice\PhpPresentation\Reader\PowerPoint2007->loadFile('/app/data/templ...')
#5 /app/vendor/phpoffice/phppresentation/src/PhpPresentation/IOFactory.php(73): PhpOffice\PhpPresentation\Reader\PowerPoint2007->load('/app/data/templ...')
```

IvanSerevko:
[...]

PR #752, opened 2023-08-14
Fixes for #710

Offbeatmammal:
creating a PR for the issue in #710 / #751


Offbeatmammal:
Further testing shows that while these fixes do allow existing presentations with Notes to be open, they are still not read correctly, and if simply using the functions to duplicate the slide deck using addExternalSlide the resulting deck is corrupt. Closing this PR as it won't resolve the issues correctly

PR #771, opened 2023-12-04
PowerPoint2077 Writer : Fixed error when defining min/max bounds to 0

Progi1984:
Superseeds #754 by @LilyEssence


Progi1984:
Thanks @LilyEssence for your contribution

PR #773, opened 2023-12-04
PowerPoint2007 Writer : Fixed column indices for embedded spreadsheets

Progi1984:
Superseeds #664 by @michael-roth


Progi1984:
Thanks @michael-roth

PR #572, opened 2020-03-11
Raised max value for identifier rand call

Scheissy:
Raised the max value for the `rand()` call so dublicated identifier are less likely. See issue [#569](https://github.com/PHPOffice/PHPPresentation/issues/569).

PR #777, opened 2023-12-06
Raised max value for identifier rand call

Progi1984:
Superseeds #572 by @Scheissy


Progi1984:
Thanks @Scheissy for your contribution

PR #778, opened 2023-12-07
PowerPoint2007 Writer : Fixed issue when first element in series is null

Progi1984:
Superseeds #671


Progi1984:
Thanks @ksmeeks0001for your contribution

PR #784, opened 2023-12-12
Samples : Allow to run without composer

Progi1984:
Superseeds #609 by @pal-software


Progi1984:
Thanks @pal-software for your contribution.

PR #793, opened 2024-02-28
Fixed fonts in the generated file after usiong a reader

ag3202:
Discovered a font issue in the scenario of importing and then exporting, not sure about other scenarios. Unclear about the handling of Panose length and adding zeros, as Panose can easily exceed 10, and not very familiar with using dechex. After fixing, the font displays correctly, placing a:latin in a:ea results in the font displaying correctly after modification.

### DIFF
--- a/src/PhpPresentation/Reader/PowerPoint2007.php
+++ b/src/PhpPresentation/Reader/PowerPoint2007.php
@@ -485,6 +485,8 @@ class PowerPoint2007 implements ReaderInterface
                         // Background
                         $oBackground = new Slide\Background\Image();
                         $oBackground->setPath($tmpBkgImg);
+                        $extension = pathinfo( $pathImage, PATHINFO_EXTENSION );
+                        $oBackground->setExtension($extension);
                         // Slide Background
                         $oSlide = $this->oPhpPresentation->getActiveSlide();
                         $oSlide->setBackground($oBackground);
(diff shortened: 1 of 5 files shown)


xlar8or:
I'm experiencing this issue, it's throwing the exception because the panose is already 020B0604020202020204 instead of 2B64222224. Instead of removing the len check we could check if it's 20 before throwing the exception, if it is, it runs a logic that cuts the padded zeros.

Progi1984:
@xlar8or Could you rebase your PR against the master branch, please? Thanks

xlar8or:
> @xlar8or Could you rebase your PR against the master branch, please? Thanks

Seems to work, thanks! I'm stil having having issues where the save of a pptx is broken (vertical text instead of horizontal text, wrong font size, shapes aren't saved, image sizes aren't the same, audio/video isn't saved, etc), so it doesn't seem to be related with panose. I've already commented on this issue in https://github.com/PHPOffice/PHPPresentation/issues/837 so hopefully you'll be able to look at that soon, because due to this issue it unfortunantely makes the library unusable to output a pptx that isn't fully made with the library functions and uses slides from another pptx.

Progi1984:
Yeah. I would like to merge this PR for fixing font issues and continue on issues next week. 

So i need you rebase your PR with the font issue fix.

xlar8or:
> Yeah. I would like to merge this PR for fixing font issues and continue on issues next week.
> 
> So i need you rebase your PR with the font issue fix.

I've replaced my entire library with the latest master and the font fix works, so it's all good on my side.

Progi1984:
@xlar8or So sorry 

@ag3202 Could you rebase it ?

Progi1984:
Superseeded by #851

PR #797, opened 2024-03-27
Data type is num unless one of the values is non-numeric

LilyEssence:
What:
- only set datatype to 'str' if one of the values is non-numeric

Why:
- if the first value is numeric, but the later values are not, for example ['1', 'FOO'], the entry will be set to numlit, but since the second entry is not a numeric, the ppt fails to open and requires repair.
- The value should be set to 'strlit' if any of the values are non-numeric.


Progi1984:
Fixed in #778 by @ksmeeks0001

PR #807, opened 2024-06-27
SetImageResource receives unsupported output

jaapdh:
SetImageResource receives the output from imagecreatetruecolor which can also be false or GdImage.

This solves the PHPStan failing on the current config.


src/PhpPresentation/Shape/Drawing/Gd.php
```diff
@@ -83,15 +85,15 @@ public function getImageResource()
     /**
      * Set image resource.
      *
-     * @param resource $value
+     * @param resource|false|GdImage|null $value
```
Progi1984, line 88: @jaapdh Could you revert these changes, please ?

Progi1984:
Superseded by #843

PR #876, opened 2025-08-05
Fix SVG errors

seanlynchwv:
If you add an svg file currently it adds 2 error notification popups to a deck


seanlynchwv:
Another PR on the way here. I will create a minimalistic PR in the future. Auto-pilot has me requesting a PR into the main repo not my custom branch

PR #885, opened 2026-01-08
Add support for text outline styling

electronic-club:
Add support for text outline styling and conditional fill rendering in fonts


electronic-club:
Bildschirmfoto 2026-01-08 um 13 44 36

PR #887, opened 2026-02-12
Refactor shadow style handling in HTML writer

phili67:
### Description

There is a bug in the case the is_null($style) and !is_null($shape->getShadow()) too

Fixes # (issue)

### Checklist:

- [ ] My CI is :green_circle:
- [ ] I have covered by unit tests my new code (check build/coverage for coverage report)
- [ ] I have updated the [changelog](https://github.com/PHPOffice/PHPPresentation/blob/develop/docs/changes/1.1.0.md)

### DIFF
--- a/src/PhpPresentation/Writer/HTML.php
+++ b/src/PhpPresentation/Writer/HTML.php
@@ -229,7 +229,9 @@ class HTML extends AbstractWriter implements WriterInterface
         $styles[] = 'height: ' . $shape->getHeight() . 'px';
         $styles[] = 'top: ' . $shape->getOffsetY() . 'px';
         $styles[] = 'left: ' . $shape->getOffsetX() . 'px';
-        $styles = array_merge($styles, $this->getStyleShadow($shape->getShadow()));
+        if (!is_null($shape->getShadow())) {
+            $styles = array_merge($styles, $this->getStyleShadow($shape->getShadow()));
+        }
 
         $this->bodySlides .= '<img src="' . $imageData . '" style="' . implode(';', $styles) . '" alt="' . $shape->getDescription() . '" title="' . $shape->getDescription() . '" />';
     }
@@ -365,7 +367,7 @@ class HTML extends AbstractWriter implements WriterInterface
      */
     protected function getStyleShadow(Shadow $style): array
     {
-        if (!$style->isVisible()) {
+        if (is_null($style) or !is_null($style) and !$style->isVisible()) {
             return [];
         }
         $boxShadow = '';


src/PhpPresentation/Writer/HTML.php
```diff
@@ -365,7 +367,7 @@ protected function getStyleFont(Font $style): array
      */
     protected function getStyleShadow(Shadow $style): array
     {
-        if (!$style->isVisible()) {
+        if (is_null($style) or !is_null($style) and !$style->isVisible()) {
```
Progi1984, line 370:
```suggestion
        if (!$style->isVisible()) {
```

Progi1984 (commented):
Hi,

Could you add an unit test? 

Thanks

dkulyk:
The bug you found here is real — I could reproduce it — but it reaches further than the HTML writer, and I think the fix belongs one level up. I have opened #928 with that shape and credited you in the changelog line; this comment is what I measured, so you can judge it for yourself.

**The trigger is the setter.** `AbstractShape::setShadow(?Shadow $pValue = null)` stores whatever it is given, and `getShadow(): ?Shadow` hands the null on. Every writer reads it without asking whether it is there, and every one declares a non-nullable `Shadow` parameter. On a shape with `setShadow(null)`:

| written as | result |
| --- | --- |
| HTML, rich text | OK — no shadow read from a text body |
| PowerPoint2007 | `TypeError: AbstractSlide::writeShadow(): Argument #2` |
| ODPresentation | `TypeError: Content::writeStylePartShadow(): Argument #2` |
| HTML, image | `TypeError: HTML::getStyleShadow(): Argument #1` |
| HTML, video | `TypeError: HTML::getStyleShadow(): Argument #1` |

Five sites, three writers: `AbstractSlide.php:273` and `:1534`, `Content.php:1011` and `:1142`, `HTML.php:232` and `:301`.

**Two things about the diff here, meant as review notes:**

1. The guard is on `writeImage()` (`:232`) but not on `writeVideo()` (`:301`), twelve lines further down — same call, same null, same TypeError.

2. The second hunk cannot fire. `getStyleShadow(Shadow $style)` keeps its non-nullable parameter, so `is_null($style)` is unreachable: PHP throws before the body runs. Untangling the precedence, `is_null($style) or !is_null($style) and !$style->isVisible()` is equivalent to the `!$style->isVisible()` that was already there — so it changes no behaviour, but it does break two checks. I applied your diff locally to confirm what CI is complaining about: PHPStan says *Call to function is_null() with PhpOffice\PhpPresentation\Style\Shadow will always evaluate to false*, and PHP CS Fixer wants `||`/`&&` rather than `or`/`and`.

**What #928 does instead:** the setter replaces a null with a fresh `Shadow`, and `getShadow()` narrows to `Shadow`. A shape is already born with `new Shadow()` in the constructor, and `Shadow` carries `isVisible()` for saying "no shadow please" — so a null was never a state worth holding. One change, all five sites correct, no guard added anywhere. It is the same treatment `setFont()` got in #924.

One more thing worth knowing about this PR's CI, unrelated to the code: seven of the fifteen red checks are `PHP Static Analysis (7.1/7.2/7.3)`, `PHPUnit 7.2/7.3` and `Check samples (7.2/7.3)` — jobs that no longer exist in the matrix since #894 raised the minimum PHP to 7.4. A rebase would clear those on its own.

dkulyk:
[#928](https://github.com/PHPOffice/PHPPresentation/pull/928) is merged (`74954c776`), so the bug you found here is fixed: a shadow set to `null` is replaced with a fresh one in `AbstractShape` rather than guarded for in the HTML writer, so all three Writers get it rather than one. Your name is on the changelog line for it.

@Progi1984 this one can be closed.

PR #860, opened 2025-03-28
feat: add basic reader and writer for keynote

sparkybug:
### Summary
Added support for Apple IWork (Keynote) format in PHPPresentation with:
- New IWork reader/writer implementation
- Snappy compression support
- Protobuf message encoding/decoding
- Added `google/protobuf` dependency

### Motivation
Enable PHPPresentation to work with Apple Keynote presentations, improving cross-platform compatibility.

### Technical Changes
- Added compression and encoding utilities
- Implemented IWork format handlers
- Added format to IOFactory registry
- Includes test coverage

This change enables users to read and write Apple Keynote presentations while maintaining PHPPresentation's existing functionality.

Fixes #48
/claim #48 

### Checklist:

- [x] My CI is :green_circle:
- [x] I have covered by unit tests my new code (check build/coverage for coverage report)
- [ ] I have updated the [documentation](https://github.com/PHPOffice/PHPPresentation/tree/develop/docs) to describe the changes
- [ ] I have updated the [changelog](https://github.com/PHPOffice/PHPPresentation/blob/develop/docs/changes/1.1.0.md)

PR #891, opened 2026-06-25
Add Keynote reader and writer

### LINKED ISSUE #48: Reader / Writer for Keynote
Ref : @surefirejack (#46)

Links : 
- https://github.com/obriensp/iWorkFileFormat/blob/master/Docs/index.md
## 

Goal : Basic Reader/Writer : Image and Text

yasumorishima:
## What it does

Adds a **Reader** and **Writer** for the Apple Keynote (`.key`) presentation format.

The implementation targets the **legacy iWork '09 structure**, where a Keynote
package is a Zip archive containing an `index.apxl` XML document. Both classes
follow the existing reader/writer conventions (`ReaderInterface` / `AbstractWriter`)
and are registered in `IOFactory` for automatic resolution.

Supported in this first milestone:

- Slides
- Text placeholders — paragraphs and runs
- Shape geometry (position and size)
- Speaker notes

## Why

Resolves the long-standing request for Keynote support (#48). A previous attempt
(#860) targeted the modern IWA binary format and stalled; this PR deliberately
scopes to the XML-based iWork '09 structure so it can be implemented in pure PHP
(no new dependencies), stays compatible with the project's supported PHP versions
(7.4+), and is fully verifiable in CI.

## Scope / limitations (intentional)

- The **modern IWA binary format** (Keynote '13 and later: Snappy-compressed
  Protobuf) is **out of scope** and is safely rejected by `canRead()` (no false
  positives).
- Because of the format generation gap, output is **not guaranteed** to open
  directly in a current Keynote.app. Correctness is instead guaranteed by a
  **round-trip test** (write → read → assert structure).
- Content that the basic format cannot represent — non-text shapes (images,
  lines, tables, charts) and in-run line breaks — is **silently skipped** on
  save; this is characterized by tests (see below).
- Styling (font/bold/size/color), themes and document properties are left as
  follow-up work.

## How it was tested

- New unit tests for the reader and writer, plus a dedicated round-trip test.
- Reader coverage includes `canRead()` rejection of non-Keynote files, non-Zip
  files and directories, and `load()` error cases (missing file, wrong format).
- Round-trip coverage includes: multi-slide content with geometry, paragraphs,
  runs and speaker notes; an empty presentation; **Unicode and XML metacharacters**
  (`< > & " '`, non-ASCII, emoji) surviving escaping unchanged; **multiple text
  placeholders per slide** preserved in order; and **non-representable content**
  (a line shape and an in-run line break) being skipped without data corruption.
- `phpunit`, `phpstan` (level 6), `php-cs-fixer` and `phpmd` all pass; the full
  existing suite remains green (no regressions).

## Notes

- No new runtime dependencies; uses only `ext-xml` and `ext-zip` (already required).
- Documentation updated in `docs/usage/readers.md` and `docs/usage/writers.md`.

Closes #48

/claim #48

### DIFF
--- a/docs/changes/1.3.0.md
+++ b/docs/changes/1.3.0.md
@@ -14,6 +14,7 @@
 - PowerPoint2007 Reader : Read a group of shapes, and map the shapes it holds out of the coordinate space `a:chOff`/`a:chExt` declares, by [@Wilkolicious](http://github.com/Wilkolicious) in [#870](https://github.com/PHPOffice/PHPPresentation/pull/870) and [@dkulyk](http://github.com/dkulyk) in [#947](https://github.com/PHPOffice/PHPPresentation/pull/947)
 - Added `Shape\RichText\Field`, so that a slide number, a slide count or a date can stand inside a sentence rather than take a whole shape, written and read back by the PowerPoint2007 Writer and Reader and by the ODPresentation Writer and Reader, by [@dkulyk](http://github.com/dkulyk) fixing [#958](https://github.com/PHPOffice/PHPPresentation/issues/958) in [#959](https://github.com/PHPOffice/PHPPresentation/pull/959)
 - PowerPoint2007 Writer & Reader : Added a fill behind the data labels of a chart serie, so that a label stays readable over the chart it sits on, by [@dkulyk](http://github.com/dkulyk) in [#963](https://github.com/PHPOffice/PHPPresentation/pull/963)
+- Keynote: Add basic reader and writer for the Apple Keynote (.key) format by [@yasumorishima](https://github.com/yasumorishima) in [#891](https://github.com/PHPOffice/PHPPresentation/pull/891)
 
 ## Bug fixes
 - PowerPoint2007 Reader : Fixed a colour with an alpha coming back opaque and one character short, taking the first digit of the colour with it, by [@dkulyk](http://github.com/dkulyk) in [#961](https://github.com/PHPOffice/PHPPresentation/pull/961)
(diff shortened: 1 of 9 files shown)


yasumorishima:
A quick note on the currently red CI jobs — they are all unrelated to this change:

- **PHPUnit / PHPStan / samples on PHP 7.2 & 7.3** fail at `composer install`: dependencies can no longer be resolved because `phpoffice/phpspreadsheet` now requires PHP `>= 7.4`. These jobs fail before running anything, and this affects the repository's CI in general on those versions (not specific to this PR).
- **PHPStan on PHP 8.4**: with no committed `composer.lock`, the latest `phpunit/phpunit` (13.x) is installed, which PHPStan then reports as `extends unknown class PHPUnit\Framework\TestCase` across all existing test classes (e.g. `AbstractShapeTest`). This is an environment/tooling issue, not something introduced here.
- **PHPUnit 8.1**: a single pre-existing failure in `Writer\ODPresentation\ObjectsChartTest::testTypeAxisUnit`, unrelated to Keynote.

The new Keynote reader/writer is green on every job where the environment resolves correctly: PHPStan 7.1 / 7.4 / 8.0–8.3, PHPUnit 7.1 / 7.4 / 8.0 / 8.2 / 8.3 / 8.4, all `samples` jobs, PHP-CS-Fixer and PHPMD.

Happy to adjust anything on request.

yasumorishima:
Rebased onto `master` now that #897 is merged.

The two `PHP Static Analysis (8.4)` / `(8.5)` failures on this PR came from PHPStan 1.x being unable to resolve PHPUnit 13 symbols on PHP 8.4/8.5, which #897 fixed by moving to PHPStan `^2.2`. After rebasing, PHPStan 2.2.8 at level 6 (no baseline) reported three issues in this PR's own files, all fixed here:

- `Writer/Keynote.php`: dropped the `'' == $pFilename` half of the `tempnam()` guard — `tempnam()` returns `non-falsy-string|false`, so the comparison was always false.
- `Tests/KeynoteRoundtripTest.php`: removed two `instanceof RichText` guards that follow an `assertInstanceOf(RichText::class, ...)` — they were always true, and calling the getters directly makes the assertions unconditional.

The presentation diff is otherwise unchanged from the previous head.

Progi1984:
Fixed in #949

PR #897, opened 2026-08-06
Fix static analysis on PHP 8.4 and 8.5

yasumorishima:
## Problem

`PHP Static Analysis (8.4)` and `(8.5)` fail on `master` as well as on every pull request, with 3451 errors of this shape:

```
Call to an undefined static method PhpOffice\PhpPresentation\Tests\Reader\PowerPoint2007Test::assertEquals().
```

The methods are not missing: in PHPUnit 13.2.6 `TestCase` still extends `Assert` and `assertEquals()` is still `final public static`. The analyser is what cannot see them — the job installs **PHPStan 1.12.34** (the last 1.12.x release, July 2025) together with **PHPUnit 13.2.6**, which composer resolves on PHP 8.4/8.5 because `phpunit/phpunit` is constrained as `>=7.0`. PHPStan 1.x cannot discover the symbols of that PHPUnit generation, and it prints its own advice in the log:

```
Upgrade today to PHPStan 2.2 or newer by using "phpstan/phpstan": "^2.2" in your composer.json.
```

The PHP 7.4–8.3 jobs stay green because an older PHPUnit is resolved there.

## Change

- `phpstan/phpstan` is raised to `^2.2`. PHPStan 2.2.8 still requires `php: ^7.4|^8.0`, so every job in the matrix keeps resolving.
- `phpstan-baseline.neon` records the findings the newer analyser reports on the existing code, so this pull request stays a CI fix and does not mix in unrelated changes. Each PHP version resolves a different PHPUnit generation (9.6.35, 10.5.64, 11.5.56, 12.5.33, 13.2.6) and therefore reports a slightly different set — for example `getMockForAbstractClass()` only exists up to PHPUnit 9 — so the baseline is the union over all of them, 98 entries. `reportUnmatchedIgnoredErrors: false` is already set, so the entries that do not apply to a given version are simply not reported.
- `ObjectsChart` now derives the chart column name from its index instead of incrementing a string. String increment is deprecated as of PHP 8.5 and was the only new finding specific to that version (`Cannot use ++ on string`).

The findings kept in the baseline are mostly redundant PHPUnit assertions (`assertIsArray()` on a value already typed `array`), `@var` PHPDoc tags that are not a subtype of the native type, and a few always-true strict comparisons. I am happy to work through them in follow-ups if you would rather see them fixed than baselined.

## Verification

Run in a container with PHP 8.4.15 and PHP 8.5.0. Older versions were reproduced by pinning `config.platform.php` so composer resolves the same dependency set as the matrix job (that pin is not part of the diff).

| PHP | PHPUnit resolved | phpstan before | phpstan after |
| --- | --- | --- | --- |
| 7.4 | 9.6.35 | 48 errors (with PHPStan 2 and no baseline) | **No errors** |
| 8.0 | 9.6.35 | same | **No errors** |
| 8.1 | 10.5.64 | same | **No errors** |
| 8.2 | 11.5.56 | same | **No errors** |
| 8.3 | 12.5.33 | — | **No errors** |
| 8.4 | 13.2.6 | 3451 errors | **No errors** |
| 8.5 | 13.2.6 | 3451 errors + `Cannot use ++ on string` | **No errors** |

Other checks, before and after the change: `phpmd src/,tests/ text ./phpmd.xml.dist` clean, `php-cs-fixer fix --dry-run --diff` 0 of 275 files, `composer validate --strict` valid, and the ODPresentation writer test suite identical at 95 tests / 1555 assertions (the one error in both columns is `MetaInfManifestTest::testMemoryDrawing` calling `imagecreatetruecolor()`, i.e. the container missing `ext-gd`).

The column-name helper was compared against the previous behaviour for every index from 2 to 2000 (`B`, `C`, …, `Z`, `AA`, …): 1999 values, 0 mismatches, including the `Z → AA` and `ZZ → AAA` boundaries.

## Note on the PHPUnit jobs

While verifying this I saw `ObjectsChartTest::testTypeAxisUnit` fail intermittently in CI with `Element automatic-styles failed to validate content`. It is not related to this change: the same commit produced both a failing and a passing run of that job, on different PHP versions and at different lines of the test, re-running the failed job on the same commit succeeded, and running the full suite three times with and three times without this change gave an identical set of failures locally. It looks like a pre-existing non-deterministic test; I have not touched it here.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
     "require-dev": {
         "phpunit/phpunit": ">=7.0",
         "phpmd/phpmd": "2.*",
-        "phpstan/phpstan": "^0.12.88 || ^1.0.0",
+        "phpstan/phpstan": "^2.2",
         "dompdf/dompdf": "^3.1"
     },
     "suggest": {
--- a/docs/changes/1.3.0.md
+++ b/docs/changes/1.3.0.md
@@ -9,4 +9,5 @@
 ## Bug fixes
 - Guarded `imagedestroy()` calls behind a PHP version check (no-op since PHP 8.0, deprecated in PHP 8.5) by [@slayerfx](http://github.com/slayerfx) in [#893](https://github.com/PHPOffice/PHPPresentation/pull/893)
 - Fixed code coverage configuration for PHPUnit 10+ by [@slayerfx](http://github.com/slayerfx) in [#895](https://github.com/PHPOffice/PHPPresentation/pull/895)
+- Fixed static analysis on PHP 8.4 and 8.5 (PHPStan 1.x could not resolve the symbols of PHPUnit 13) by [@yasumorishima](http://github.com/yasumorishima) in [#897](https://github.com/PHPOffice/PHPPresentation/pull/897)
 
--- a/src/PhpPresentation/GeometryCalculator.php
+++ b/src/PhpPresentation/GeometryCalculator.php
@@ -37,7 +37,7 @@ class GeometryCalculator
     {
         $offsets = [self::X => 0, self::Y => 0];
 
-        if (null !== $container && 0 != count($container->getShapeCollection())) {
+        if (0 != count($container->getShapeCollection())) {
             $shapes = $container->getShapeCollection();
             if (null !== $shapes[0]) {
                 $offsets[self::X] = $shapes[0]->getOffsetX();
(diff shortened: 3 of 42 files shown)


Progi1984 (commented):
I'm not a fan of all ignoring messages. I prefer fixes.

Progi1984 (approved):

Issue #705, opened 2022-04-06
Mark shape as Decorative


OrigoPerry:
To improve accessibility it would be great if shapes could be marked as decorative. A true/false/null property that if not null would add the following to the xml:

```
<a:ext uri="{C183D7F6-B498-43B3-948B-1728B52AA6E4}">
     <adec:decorative xmlns:adec="http://schemas.microsoft.com/office/drawing/2017/decorative" val="0"/>
</a:ext>

```

PR #898, opened 2026-08-17
Expose an alternative text on every shape

dkulyk:
### Description

Only `Shape\AbstractGraphic` (and, through it, `Chart` and `Table`) carried a description, and it was written by the PowerPoint2007 writer alone. Text boxes, auto shapes, groups and lines had no way to tell assistive technologies what they are, so a presentation built from them cannot satisfy WCAG 1.1.1 — and nothing at all was exported to ODPresentation.

This moves `description` up from `Shape\AbstractGraphic` to `AbstractShape` and makes both writers and both readers handle it.

```php
$slide->createRichTextShape()
    ->setName('Budget')
    ->setDescription('Budget spent to date: 45% of 1.2M EUR');
```

- **PowerPoint2007**: the `descr` attribute of `p:cNvPr`, now written for `p:sp` (rich text and auto shapes), `p:cxnSp` (lines) and `p:grpSp` (groups) as it already was for pictures, tables and charts. `writeShapeAutoShape()` used to write a hard-coded `descr=""`, and its `name` too — both now come from the shape.
- **ODPresentation**: the `svg:desc` child element, written for frames (text, drawing, media, table, chart), lines and groups. It is placed where each ODF content model expects it: first in `draw:line` and `draw:g`, last in `draw:frame`, so `assertIsSchemaOpenDocumentValid('1.2')` passes.
- **Readers**: `Reader\PowerPoint2007::loadShapeRichText()` now reads `descr` like the other loaders; `Reader\ODPresentation` reads `svg:desc`, falling back to `draw:name` so files written by earlier versions keep their behaviour.

This is accessibility work: a shape without an alternative text is unreachable for a screen
reader, which is what WCAG 1.1.1 asks about. It closes no issue on its own, and is related to
#705 — a description says what a shape means, the decorative flag asked for there says it means
nothing; #899 covers that side.

### Checklist:

- [x] My CI is :green_circle:
      `AbstractShapeTest::testDescription`, `PptSlidesTest::testShapeDescription` (schema-validated against ECMA-376), `ContentTest::testShapeDescription` and `testShapeDescriptionOmittedWhenEmpty` (schema-validated against ODF 1.2), and a save and load round trip through both readers in `Reader\PowerPoint2007Test::testShapeDescription` and `Reader\ODPresentationTest::testShapeDescription`, the latter covering the fall back to `draw:name`.
      An *Alternative text* section in `docs/usage/shapes/introduction.md`.


Progi1984 (commented):
With current readers, can't you add assertions for unit tests ?

dkulyk:
Good point — added in 4c8f5e9.                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                     
  `Reader\PowerPoint2007Test::testShapeDescription` and                                                                                                                                                                                                                                                              
  `Reader\ODPresentationTest::testShapeDescription` save a presentation and                                                                                                                                                                                                                                          
  load it back, asserting the description survives the round trip for every                                                                                                                                                                                                                                          
  shape kind the readers load: rich text for PowerPoint2007, rich text and                                                                                                                                                                                                                                           
  drawing for ODPresentation. The ODPresentation one also covers a shape                                                                                                                                                                                                                                             
  with a `draw:name` and no `svg:desc`, ie a file written by an earlier                                                                                                                                                                                                                                              
  version, which falls back to the name.

Progi1984 (approved):

Progi1984:
@dkulyk Thank you for your contribution

PR #899, opened 2026-08-17
Mark a shape as decorative

### LINKED ISSUE #705: Mark shape as Decorative

dkulyk:
### Description

A shape that carries no information — a coloured band, a rule, a background image — is announced by assistive technologies like any other, which walks blind users through content that means nothing. Both formats can say otherwise, and the library could not.

`AbstractShape` gains a `decorative` flag, false by default. Only a decorative shape writes anything: false and absent mean the same thing to both formats.

```php
$slide->createLineShape(10, 10, 100, 10)->setDecorative();
```

- **PowerPoint2007**: the `{C183D7F6-B498-43B3-948B-1728B52AA6E4}` extension of `p:cNvPr`, written for every shape kind. Pictures may already write an `a:extLst` for the SVG creation id, so the two extensions share one list rather than emitting a second one.
- **ODPresentation**: the `loext:decorative` attribute of the shape element. ODF has no such attribute before 1.4, so this is the LibreOffice extension — which is why `ContentTest::testShapeDecorative` asserts `assertIsSchemaOpenDocumentNotValid('1.2')`, the same way the existing `loext:mime-type` test does.
- **Readers**: both restore the flag, so a load/save round trip keeps it. `Reader\PowerPoint2007` reads the extension of `p:cNvPr` for rich text, pictures, tables and charts (the shape kinds it loads), `Reader\ODPresentation` reads `loext:decorative` off the frame.

Fixes #705.

### Checklist:

- [x] My CI is :green_circle:
      `AbstractShapeTest::testDecorative`, `PptSlidesTest::testShapeDecorative` (schema-validated against ECMA-376), `ContentTest::testShapeDecorative`, and a load/save round trip in `Reader\PowerPoint2007Test::testShapeDecorative` and `Reader\ODPresentationTest::testShapeDecorative`. The `adec` prefix is registered in `PhpPresentationTestCase` so it can be reached by XPath.
      A *Decorative shapes* section in `docs/usage/shapes/introduction.md`.

### DIFF
--- a/src/PhpPresentation/AbstractShape.php
+++ b/src/PhpPresentation/AbstractShape.php
@@ -126,6 +126,14 @@ abstract class AbstractShape implements ComparableInterface
     protected $description = '';
 
     /**
+     * Decorative, ie the shape is ignored by assistive technologies.
+     * Null when the document says nothing about it.
+     *
+     * @var null|bool
+     */
+    protected $decorative;
+
+    /**
      * Create a new self.
      */
     public function __construct()
@@ -241,6 +249,28 @@ abstract class AbstractShape implements ComparableInterface
     public function setDescription(string $pValue = ''): self
     {
         $this->description = $pValue;
+
+        return $this;
+    }
+
+    /**
+     * Is the shape decorative, ie ignored by assistive technologies?
+     * Null when the document says nothing about it.
+     */
+    public function isDecorative(): ?bool
+    {
+        return $this->decorative;
+    }
+
+    /**
+     * Set the shape as decorative, ie ignored by assistive technologies.
+     * Null removes the information from the document.
+     *
+     * @return static
+     */
+    public function setDecorative(?bool $pValue = true): self
+    {
+        $this->decorative = $pValue;
 
         return $this;
     }
--- a/src/PhpPresentation/AbstractShape.php
+++ b/src/PhpPresentation/AbstractShape.php
@@ -127,11 +127,10 @@ abstract class AbstractShape implements ComparableInterface
 
     /**
      * Decorative, ie the shape is ignored by assistive technologies.
-     * Null when the document says nothing about it.
      *
-     * @var null|bool
+     * @var bool
      */
-    protected $decorative;
+    protected $decorative = false;
 
     /**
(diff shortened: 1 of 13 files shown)


src/PhpPresentation/AbstractShape.php
```diff
@@ -118,6 +118,14 @@ abstract class AbstractShape implements ComparableInterface
      */
     protected $name = '';
 
+    /**
+     * Decorative, ie the shape is ignored by assistive technologies.
+     * Null when the document says nothing about it.
+     *
+     * @var null|bool
```
Progi1984: I think you can define as a bool. With false per default.

Progi1984 (changes requested):

src/PhpPresentation/AbstractShape.php
dkulyk: done, now a plain bool

Progi1984 (approved):

Progi1984:
@dkulyk Can you rebase and fix conflicts ?

dkulyk:
@Progi1984 Rebased onto `master`, conflicts resolved — they were all with #898, which touches the same call sites.

Two orderings mattered:
- **PowerPoint2007**: `descr` is an attribute of `p:cNvPr`, the decorative flag is its `a:extLst` child, so the attribute is written first.
- **ODPresentation**: the other way round — `loext:decorative` is an attribute of the shape element, `svg:desc` is a child, so the flag goes first.

`Reader\ODPresentation::loadShapeDrawing` now takes the description from `loadShapeDescription()` instead of falling back to `draw:name` itself. Tests for both features are kept as separate methods.

Progi1984 (approved):

Progi1984:
@dkulyk Thank you for your contribution 🥰

PR #901, opened 2026-08-17
ODPresentation Writer : Fix the chart output

dkulyk:
### Description

A chart written by the `ODPresentation` writer does not come out of LibreOffice looking like the chart it was given. The same presentation, one bar serie of four values with the second one filled red, written before and after this pull request :

| Before | After |
| --- | --- |
| before | after |

Five defects are at work in that picture, and this pull request fixes them one by one. The reference used throughout is a chart LibreOffice writes itself : insert a default chart, save as `.odp`, read `Object 1/content.xml`.

**1. A serie drew a single data point.** `writeSeries()` emitted one `chart:data-point chart:repeated="1"` for the whole serie, so a serie of four values was drawn as one bar, and a styled point shifted the values along. The points taking the style of the serie are now counted and their run written with the right `chart:repeated`, the styled ones written on their own — which is what LibreOffice writes :

```xml
<chart:series chart:style-name="ch7" …><chart:data-point chart:repeated="4"/></chart:series>
```

**2. Every chart came out on a black background.** `writeChartStyle()` wrote `draw:fill-color` without `draw:fill`. The ODF default for `draw:fill` is `solid`, so the colour of a `Fill` that was never asked to be painted was taken as a solid background. `draw:fill` is now written explicitly, as LibreOffice does :

```xml
<style:style style:name="ch1" …><style:graphic-properties draw:stroke="none" draw:fill="none"/></style:style>
```

**3. A hidden legend was drawn.** `writeLegend()` never looked at `Legend::isVisible()`. The element is now left out when the legend is hidden, absence of `chart:legend` being how ODF says a chart has no legend.

**4. The data labels never showed.** `chart:data-label-number` was written on the style of the serie only, where LibreOffice reads it from the style of the plot area. The format asked for by the first serie that shows something is now written there as well, `percentage` being forced for a percent-stacked bar.

**5. A serie with no fill of its own came out transparent.** `writeSeriesStyle()` wrote `draw:fill` from `$series->getFill()->getFillType()`, and the default `Fill` of a serie is `FILL_NONE` — so a serie whose colour was never set asked ODF for no fill at all. A serie leaving its fill untouched now says nothing about its colour and the application picks one, which is what the `PowerPoint2007` writer already does (`PptCharts.php`, `Fill::FILL_NONE != $series->getFill()->getFillType()`) and what LibreOffice does with its own palette :

```xml
ch7: draw:stroke="none" draw:fill-color="#004586"
ch8: draw:stroke="none" draw:fill-color="#ff420e"
```

As a side effect of the same change, `Series::setFill(null)` no longer fatals in this writer.

Nothing here touches the `PowerPoint2007` writer : its output was already correct in all five cases, and the tests cover the ODF side only.

### Checklist:

- [x] My CI is :green_circle:
- [x] I have updated the [documentation](https://github.com/PHPOffice/PHPPresentation/tree/develop/docs) to describe the changes — no public API changes, the documented behaviour is the one that was broken

### DIFF
--- a/docs/changes/1.3.0.md
+++ b/docs/changes/1.3.0.md
@@ -12,4 +12,5 @@
 - Guarded `imagedestroy()` calls behind a PHP version check (no-op since PHP 8.0, deprecated in PHP 8.5) by [@slayerfx](http://github.com/slayerfx) in [#893](https://github.com/PHPOffice/PHPPresentation/pull/893)
 - Fixed code coverage configuration for PHPUnit 10+ by [@slayerfx](http://github.com/slayerfx) in [#895](https://github.com/PHPOffice/PHPPresentation/pull/895)
 - Fixed static analysis on PHP 8.4 and 8.5 (PHPStan 1.x could not resolve the symbols of PHPUnit 13) by [@yasumorishima](http://github.com/yasumorishima) in [#897](https://github.com/PHPOffice/PHPPresentation/pull/897)
+- ODPresentation Writer : Fixed the charts losing their data points, their background, their data labels and the colour of their series, and drawing a hidden legend by [@dkulyk](http://github.com/dkulyk) in [#901](https://github.com/PHPOffice/PHPPresentation/pull/901)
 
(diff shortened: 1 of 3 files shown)


Progi1984 (approved):

Progi1984:
@dkulyk Thanks for the contribution. It's clean (explanation and unit tests)

PR #883, opened 2025-12-17
Add missing slide names

potofcoffee:
Fixes #882

### Description

This includes the missing slide name attribute in PowerPoint2007 writer. No docs update necessary, since this should have been there all along.

Fixes #882 

### Checklist:

- [X] My CI is :green_circle:
- [ ] I have covered by unit tests my new code (check build/coverage for coverage report)
- [ ] I have updated the [documentation](https://github.com/PHPOffice/PHPPresentation/tree/develop/docs) to describe the changes
- [X] I have updated the [changelog](https://github.com/PHPOffice/PHPPresentation/blob/develop/docs/changes/1.1.0.md)

PR #902, opened 2026-08-18
PowerPoint2007 Writer : Keep the name of a slide

### LINKED ISSUE #882: PowerPoint2007 writer does not set slide names
The PowerPoint2007 writer does not set the name attribute for cSld elements. This means names set through $slide->setName() are lost.

dkulyk:
### Description

`Slide::setName()` had no effect on a PowerPoint2007 file. The ODPresentation writer writes it as `draw:name` on `draw:page` and `Reader\ODPresentation` reads it back, but `Writer\PowerPoint2007` never wrote anything, so the name was lost on save and there was nothing left for the reader to restore.

```php
$slide->setName('Introduction');
```

- **Writer**: the `name` attribute of `p:cSld`, written only when a name is set — `null` leaves the attribute out, the same condition the ODPresentation writer uses.
- **Reader**: `Reader\PowerPoint2007` reads it off `/p:sld/p:cSld`, so a load/save round trip keeps it. `Reader\ODPresentation` already did.

Fixes #882.

This supersedes #883 by @potofcoffee, which found the same gap. Two differences: that PR guards with `!empty()`, which also drops the legitimate name `'0'`, and it stops at the writer, so a name written by PowerPoint is still lost when the file is read back. Please close #883 when this one lands.

### Checklist:

- [x] My CI is :green_circle:
      `PptSlidesTest::testSlideName` (schema-validated against ECMA-376, mirroring the existing `ODPresentation\ContentTest::testSlideName`) and a load/save round trip in `Reader\PowerPoint2007Test::testSlideName`.
      The *Name* section of `docs/usage/slides/introduction.md` now says how the name is stored in each format.


Progi1984 (approved):

Progi1984:
@dkulyk Thank you again for your contribution 😻

dkulyk:
This is the last one for now, but I'll very likely be back when the customer requires some other enhancement in PHPPresentation :)
Thank you for applying the PRs so quickly!

Issue #610, opened 2020-12-23
Bar charts not formatting series data as numbers


Marcote2014:
Hi, I have created a series and I want to format it as a percentage and the% does not appear.
From what I have verified it is putting the formatCode label where it does not correspond.

My code:
      // Create a bar chart
      $barChart = new Bar();
      $barChart->setGapWidthPercent(50);

      $series1 = new Series('MOB SR', $series1Data);
      $series1->setShowSeriesName(FALSE);
      $series1->getFont()->getColor()->setRGB('000000');
      $series1->getFill()->setFillType(Fill::FILL_SOLID)->setStartColor(new Color('FF231161'));
      $series1->setLabelPosition(Series::LABEL_OUTSIDEEND);
      $series1->setShowPercentage(TRUE);
      // Data Label Number Format
      $series1->setDlblNumFormat('0%');

It would have to be modified for the following in the PptCharts.php file:

protected function writeMultipleValuesOrReference($objWriter, $isReference, $values, $reference, $formatCode = NULL)
    {
        // c:strLit / c:numLit
        // c:strRef / c:numRef
        $referenceType = ($isReference ? 'Ref' : 'Lit');
        $dataType = 'str';
        if (is_int($values[0]) || is_float($values[0])) {
            $dataType = 'num';
        }
        $objWriter->startElement('c:' . $dataType . $referenceType);
		
+		if (($dataType.$referenceType == 'numLit') && (!is_null($formatCode))) {
+			//c:formatCode			
+			$objWriter->writeElement('c:formatCode', $formatCode);
+		}

$numValues = count($values);
        if (!$isReference) {
...

protected function writeTypeBar(XMLWriter $objWriter, Bar $subject, $includeSheet = false)
    {
...
- if ($series->hasDlblNumFormat()) {
 -               //c:numFmt
 -               $objWriter->startElement('c:numFmt');
 -               $objWriter->writeAttribute('formatCode', $series->getDlblNumFormat());
 -               $objWriter->writeAttribute('sourceLinked', '0');
 -               $objWriter->endElement();
- }
...
+ $formatCode = ($series->hasDlblNumFormat()) ? $series->getDlblNumFormat() : NULL;
- $this->writeMultipleValuesOrReference($objWriter, $includeSheet, $axisYData, $coords);
+ $this->writeMultipleValuesOrReference($objWriter, $includeSheet, $axisYData, $coords, $formatCode);

[...]

Issue #420, opened 2017-11-16
Line charts not formatting series data as numbers


albertnel:
Hi, it seems like series data that I give to a Line() object does not format the values into numbers as they should be.

Because of this the following does not work:

```
$lineChart = new Line();

$seriesData = array('Monday' => 18, 'Tuesday' => 23, 'Wednesday' => 14, 'Thursday' => 12, 'Friday' => 20, 'Saturday' => 8, 'Sunday' => 10);
$series = new Series('example', $seriesData);
$series->setShowValue(false);
$series->setShowPercentage(true); // This does nothing
$series->setDlblNumFormat('0.00%'); // This does nothing

$lineChart->addSeries($series); 
```

Also, the Y-axis labels aren't displaying as percentages either. Can you please help or make any suggestion?

When I open the generated .pptx file and right-click on the chart and select "Edit Data", I can see that the way it's saved in Excel is of type "General" which is probably exactly why this is happening. If I change it to "Number" everything starts working.

Thanks

pascalzoet:
Hi, I have a question for you. I'm using phppresentation with laravel 5.1 and i'm trying to make charts in my powerpoint. The problem i'm facing is that whatever i try, my powerpoint will always be broken and needs repairing. 

I wondered if you you had the same problem and if you had you could tell me how to solved this

Thanks

albertnel:
Hi @pascalzoet ,

I had the exact same problem and couldn't get anything to really work with the current dev-master branch. I switched to 0.8.0 and most things started working for me. Try it and let us know.

pascalzoet:
Hi @albertnel 

Switched to version 0.8.0 and got the charts (finally) working. Thanks for the help!

albertnel:
@pascalzoet awesome, glad you got it working! :)

Progi1984:
@albertnel @pascalzoet Thanks for answers. So that works with 0.8.0 but not with the develop branch ?

pascalzoet:
@Progi1984 So i tested this code on both version 0.8.0 and dev-master

        $currentSlide = $objPHPPresentation->createSlide();
        $seriesData = array(
            'Monday' => 12,
            'Tuesday' => 15,
            'Wednesday' => 13,
            'Thursday' => 17,
            'Friday' => 14,
            'Saturday' => 9,
            'Sunday' => 7
        );

        $lineChart = new \PhpOffice\PhpPresentation\Shape\Chart\Type\Line();
[...]

PR #903, opened 2026-08-20
PowerPoint2007 Writer : Honour the format code of an axis

### LINKED ISSUE #632: Formatting axis labels with format code
I am trying to format the labels on the y axis

`$axis->setFormatCode('#.0\%');`

I have tested and this will work as expected if you change Writer\PowerPoint2007\PptCharts.php on line 2279 to 

`$objWriter->writeAttribute('sourceLinked', '0');`

Can anyone tell me why that is?

dkulyk:
### Description

`writeAxis()` writes the axis' format code and then immediately tells the renderer to ignore it:

```php
// src/PhpPresentation/Writer/PowerPoint2007/PptCharts.php:2466
$objWriter->startElement('c:numFmt');
$objWriter->writeAttribute('formatCode', $oAxis->getFormatCode());
$objWriter->writeAttribute('sourceLinked', '1');   // ← take the format from the source data
$objWriter->endElement();
```

`sourceLinked="1"` means "the number format comes from the source data", so the `formatCode` sitting next to it is discarded by PowerPoint. The attribute *is* in the file, which is why the issue reads as "it still doesn't work" rather than "nothing is written".

**Why `'0'` is the right value.** The library already writes it correctly everywhere else — in the same file, all three data-label sites use `sourceLinked="0"`:

| line | what | `sourceLinked` |
|------|------|----------------|
| 877 | data labels (`getDlblNumFormat()`) | `0` |
| 1305 | data labels | `0` |
| 1445 | data labels | `0` |
| **2468** | **axis** (`getFormatCode()`) | **`1`** |

The axis is the only one out of line. It has been hardcoded to `'1'` since b957fb9df (2016-03-02, the Gridlines implementation) and was never revisited.

**The fix** is not an unconditional `'0'` — that would take the inherited format away from every deck that never called the setter. `Axis::DEFAULT_FORMAT_CODE` is `'general'`, so:

```php
$objWriter->writeAttribute('sourceLinked', Chart\Axis::DEFAULT_FORMAT_CODE === $oAxis->getFormatCode() ? '1' : '0');
```

- an axis nobody configured → byte-identical output to before;
- an axis with an explicit format code → the format code is finally honoured.

Fixes #632

**Possibly related, deliberately not claimed here.** #610 and #420 report chart *series* data not being formatted as numbers. They may share this cause, but I have not measured them.

The ODPresentation writer never reads `getFormatCode()` at all. ODF expresses a number format as a `<number:number-style>` element referenced through `style:data-style-name`, not as an attribute, and there is no such machinery in that writer today — that is separate work, not a matter of copying this line across. Hence the `Available only on the PowerPoint2007 Writer` note in the documentation, matching the existing one on *Crossing*.

### Checklist:

- [x] My CI is :green_circle:
      `Axis::setFormatCode()` had no writer-side coverage at all — `AxisTest` only exercises the getter/setter, and `numFmt` appeared nowhere in the writer tests. `PptChartsTest::testAxisFormatCode` now covers both branches (default and explicit format code) and validates each against the ECMA-376 schema.
      `setFormatCode` was not mentioned anywhere in `docs/`. Added a *Number format* section to `docs/usage/shapes/chart.md`, next to *Bounds* and *Crossing*.

### DIFF
--- a/docs/changes/1.3.0.md
+++ b/docs/changes/1.3.0.md
@@ -34,6 +34,7 @@
 - `Style\Color` : Fixed a colour written as a plain RGB being given an alpha it never asked for by [@dkulyk](http://github.com/dkulyk) in [#910](https://github.com/PHPOffice/PHPPresentation/pull/910)
 - PowerPoint2007 Reader : Fixed a slide size written with an explicit `type="custom"` crashing the reader instead of being read with its own dimensions by [@SAY-5](http://github.com/SAY-5) in [#929](https://github.com/PHPOffice/PHPPresentation/pull/929)
 - ODPresentation Reader : Fixed the alignment of a paragraph being dropped on load, and its direction being read as left to right whatever the file said, by [@dkulyk](http://github.com/dkulyk) in [#927](https://github.com/PHPOffice/PHPPresentation/pull/927)
+- PowerPoint2007 Writer : Fixed `Axis::setFormatCode()` having no effect, the number format of an axis being taken from the source data instead by [@dkulyk](http://github.com/dkulyk) in [#903](https://github.com/PHPOffice/PHPPresentation/pull/903)
 
 ## BC Breaks
 - `\PhpOffice\PhpPresentation\Slide\SlideMaster` is constructed without a background. A deck that relied on the white fill it used to write sets one with `setBackground()`, as [the documentation](https://github.com/PHPOffice/PHPPresentation/blob/develop/docs/usage/slides/introduction.md) shows.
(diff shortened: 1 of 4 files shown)


Progi1984 (approved):
Need a rebase

dkulyk:
@Progi1984 Already rebased — `f4b42f1b3` sits directly on `307158f03` (#927), nothing behind. The push landed a few minutes before your review, so the page may have been showing the older head.

Progi1984 (approved):

PR #904, opened 2026-08-20
Let a table say whether its first row is a header row

dkulyk:
### Description

`AbstractSlide::writeShapeTable()` stamps both table-property flags on every table it writes, unconditionally:

```php
// src/PhpPresentation/Writer/PowerPoint2007/AbstractSlide.php:427
$objWriter->startElement('a:tblPr');
$objWriter->writeAttribute('firstRow', '1');
$objWriter->writeAttribute('bandRow', '1');
$objWriter->endElement();
```

A table has no say in it. For a grid with column labels that is a gift — the labels become a header row for free. For a single-column list it is wrong: the first item is announced as the heading of its column, and there is no way to turn it off short of not using a table at all.

`Table::setFirstRow()` and `Table::setBandRow()` expose both. They default to `true`, which is the behaviour the writer has always had, so a table nobody configures produces a byte-identical file.

```php
$tableShape = $slide->createTableShape($columns);
$tableShape->setFirstRow(false);   // the first row holds data, not column labels
$tableShape->setBandRow(false);
```

The reader picks the two attributes up as well (accepting `1` and `true`, matching the idiom already used at `Reader/PowerPoint2007.php:829`), so the choice survives a load/save round trip.

**One behaviour change worth naming.** A deck read from a file that carries no `a:tblPr` used to be written back with both flags on; it is now written back the way the source had it. That makes the round trip faithful, but it is a change, not just an addition.

**Deliberately out of scope.** `bandRow` stays PowerPoint2007-only — ODF expresses banding by styling each row, not as a property of the table. `firstRow` does have an ODF counterpart (`table:table-header-rows`, which the ODPresentation writer does not emit today); that is a change to the structure of every table it writes and belongs in its own PR rather than riding along here. The other four `CT_TableProperties` flags (`firstCol`, `lastRow`, `lastCol`, `bandCol`) are left alone — this PR only makes configurable what the writer already writes.

### Checklist:

- [x] My CI is :green_circle:
      `TableTest::testFirstRow` and `TableTest::testBandRow` for the shape, `PptSlidesTest::testTableWithFirstRowAndBandRow` for the writer (both branches, each validated against the ECMA-376 schema), and `Reader\PowerPoint2007Test::testTableFirstRowAndBandRow` for a load/save round trip.
      New *Header row and banded rows* section in `docs/usage/shapes/table.md`.

### DIFF
--- a/src/PhpPresentation/Shape/Table.php
+++ b/src/PhpPresentation/Shape/Table.php
@@ -44,6 +44,20 @@ class Table extends AbstractGraphic implements ComparableInterface
     private $columnCount = 1;
 
     /**
+     * Is the first row of the table styled as a header row?
+     *
+     * @var bool
+     */
+    protected $firstRow = true;
+
+    /**
+     * Are the rows of the table styled with alternating bands?
+     *
+     * @var bool
+     */
+    protected $bandRow = true;
+
+    /**
      * Create a new \PhpOffice\PhpPresentation\Shape\Table instance.
      *
      * @param int $columns Number of columns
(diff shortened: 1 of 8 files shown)


src/PhpPresentation/Shape/Table.php
```diff
@@ -43,6 +43,20 @@ class Table extends AbstractGraphic implements ComparableInterface
      */
     private $columnCount = 1;
 
+    /**
+     * Is the first row of the table styled as a header row?
+     *
+     * @var bool
+     */
+    private $firstRow = true;
```
Progi1984:
```suggestion
    protected $firstRow = true;
```

src/PhpPresentation/Shape/Table.php
```diff
@@ -43,6 +43,20 @@ class Table extends AbstractGraphic implements ComparableInterface
...
+
+    /**
+     * Are the rows of the table styled with alternating bands?
+     *
+     * @var bool
+     */
+    private $bandRow = true;
```
Progi1984:
```suggestion
    protected $bandRow = true;
```

dkulyk:
@Progi1984 Both applied.

Should `$rows` and `$columnCount` become `protected` too? They are the other two properties of the class, and `Table\Row` and `Table\Cell` are `private` throughout as well — happy to make the whole set consistent if that is the direction you want.

Progi1984 (approved):

PR #905, opened 2026-08-20
ODPresentation Writer : Write the header row of a table

dkulyk:
> **Depends on #904.** That PR adds `Table::isFirstRow()`, which this one reads. The diff shown here includes its commit until it lands; review only `ODPresentation/Content.php` and `ODPresentation/ContentTest.php`.

### Description

`Content::writeShapeTable()` writes every row as a plain `table:table-row`:

```php
// src/PhpPresentation/Writer/ODPresentation/Content.php:763
foreach ($arrayRows as $keyRow => $shapeRow) {
    // table:table-row
    $objWriter->startElement('table:table-row');
```

ODF marks a header by wrapping the rows that make it up in `<table:table-header-rows>`, and nothing did — so a table's column labels arrive as ordinary cells and are announced as such.

**Measured**, on a six-slide report written both ways and converted to PDF through LibreOffice:

| written as | `/TH` | `/TD` |
| --- | --- | --- |
| PPTX | 22 | 76 |
| ODP | **0** | 98 |

Every heading demoted. LibreOffice's own ODF export of that same deck carries `table:table-header-rows`, so the format was never the limitation.

### The change

The first row goes inside `table:table-header-rows` when the table says it is a header row — `Table::isFirstRow()`, the same flag the PowerPoint2007 writer turns into `firstRow` on the table properties. The remaining rows follow the wrapper as siblings, which is how ODF expresses it and what `assertIsSchemaOpenDocumentValid('1.2')` confirms in both branches of the new test.

`bandRow` has no counterpart here and is not touched: ODF has no table-level banding, it is a matter of styling each row.

**This changes the structure of every table the writer produces**, since a header row is the default. Three existing tests walked straight from `table:table` to `table:table-row`; their XPaths now walk through the wrapper. Only the paths moved — no assertion was changed or dropped.

### Checklist:

- [x] My CI is :green_circle:
      `ContentTest::testTableFirstRow` covers both branches — a two-row table with a header row and the same table without one — each validated against the ODF 1.2 schema.
      The *Header row and banded rows* section of `docs/usage/shapes/table.md` (added in #904) now says what each writer emits, and its PowerPoint2007-only warning is narrowed to `setBandRow` alone.


dkulyk:
Left as a draft on purpose, and it is not waiting on review feedback.

This branch is stacked on #904: its base is `feature/table-first-row`, so the two commits above are that PR's and this one's. GitHub can only open it against `master`, which is why both show here.

The reason is not just tidiness. The one line this PR turns on is `$shape->isFirstRow()` — the accessor #904 adds. On `master` alone the writer would be calling a method that does not exist.

The moment #904 is merged I will rebase this onto `master` and take it out of draft; the diff then becomes the single commit `ODPresentation Writer : Write the header row of a table`. Nothing else is outstanding — CI is green and it is ready to read now if you would rather review it before then.

dkulyk:
#904 is in, so this is rebased onto `master` and out of draft. The diff is now the single commit it was always meant to be — `ODPresentation/Content.php` and its test.

Progi1984 (approved):

PR #906, opened 2026-08-20
PowerPoint2007 Writer : Keep the borders a cell asked for

### LINKED ISSUE #808: how to border left and border bottom in powerpoint2007?
how to border left and border bottom in powerpoint2007?

thank you

dkulyk:
### Description

Fixes #808

`AbstractSlide::writeShapeTable()` hands the edge shared by two neighbouring cells to one of them, which is sound. But it inspects the **opposite side of the neighbour** from the one it then copies:

```php
// src/PhpPresentation/Writer/PowerPoint2007/AbstractSlide.php:530
if ($hasNextCellRight) {
    $nextCellRight = $shape->getRow($row)->getCell($cell + 1);
    if ($nextCellRight->getBorders()->getRight()->getHashCode() != $defaultBorder->getHashCode()) {
        $borderRight = $nextCellRight->getBorders()->getLeft();   // guard tested RIGHT, copies LEFT
    }
}
if ($hasNextRowBelow) {
    $nextCellBelow = $shape->getRow($row + 1)->getCell($cell);
    if ($nextCellBelow->getBorders()->getBottom()->getHashCode() != $defaultBorder->getHashCode()) {
        $borderBottom = $nextCellBelow->getBorders()->getTop();   // guard tested BOTTOM, copies TOP
    }
}
```

So a cell loses the borders it owns as soon as it has a neighbour on that side, and the rule that replaces them is whatever the neighbour has on the facing side — usually the default. Left and top are never overridden, which is exactly the "left works, bottom does not" shape of the report.

### Measured

A two-column, three-row table where **every** cell sets its own right and bottom rule (`LINE_THICKTHIN` / `LINE_DOUBLE`):

| cell | `a:lnR` before | `a:lnB` before | after |
| --- | --- | --- | --- |
| r0c0 | `sng` | `sng` | `thickThin` / `dbl` |
| r0c1 | `thickThin` | `sng` | `thickThin` / `dbl` |
| r1c0 | `sng` | `sng` | `thickThin` / `dbl` |
| r1c1 | `thickThin` | `sng` | `thickThin` / `dbl` |
| r2c0 | `sng` | `dbl` | `thickThin` / `dbl` |
| r2c1 | `thickThin` | `dbl` | `thickThin` / `dbl` |

Eight of the twelve rules were dropped. Only the last column kept its right border and only the last row its bottom one — the two positions with no neighbour to be overridden by.

### The fix

Inspect the side that is copied, the one facing this cell. The neighbour still wins the shared edge, but only when the neighbour is the one that asked for it — verified separately: a right neighbour that sets only its **left** border, and a cell below that sets only its **top**, both land on the top left cell as before.

Worth noting for anyone who worked around this: the old code required setting *both* sides of the neighbour — its right (to trip the guard) and its left (to supply the value). Setting just the facing side now works, and setting the border on the cell that owns it works too.

### Checklist:

- [x] My CI is :green_circle:
      The existing `testTableWithBorder` passes unchanged, which is the problem — its XPath `.../a:tr/a:tc/a:tcPr` matches any cell in any row, so default borders satisfied it. `PptSlidesTest::testTableWithBorderOnEveryCell` and `testTableWithBorderOwnedByTheNeighbour` anchor on `a:tr[1]/a:tc[1]`, the one cell that has both a neighbour to its right and a row below. Both fail on the current code and pass on the fix, and both are schema-validated.
      `docs/usage/shapes/table.md` said nothing about borders at all. New *Define the borders* section, including the sharing rule — which is what tells a caller which cell to set.

### DIFF
--- a/docs/changes/1.3.0.md
+++ b/docs/changes/1.3.0.md
@@ -38,6 +38,7 @@
 - ODPresentation Reader : Fixed the alignment of a paragraph being dropped on load, and its direction being read as left to right whatever the file said, by [@dkulyk](http://github.com/dkulyk) in [#927](https://github.com/PHPOffice/PHPPresentation/pull/927)
 - PowerPoint2007 Writer : Fixed `Axis::setFormatCode()` having no effect, the number format of an axis being taken from the source data instead by [@dkulyk](http://github.com/dkulyk) in [#903](https://github.com/PHPOffice/PHPPresentation/pull/903)
 - Fixed the fill of a table row never reaching the file, in both Writers, by [@dkulyk](http://github.com/dkulyk) in [#911](https://github.com/PHPOffice/PHPPresentation/pull/911)
+- PowerPoint2007 Writer : Fixed a cell losing its own right and bottom borders whenever it had a neighbour on that side by [@dkulyk](http://github.com/dkulyk) in [#906](https://github.com/PHPOffice/PHPPresentation/pull/906)
 
 ## BC Breaks
 - `\PhpOffice\PhpPresentation\Slide\SlideMaster` is constructed without a background. A deck that relied on the white fill it used to write sets one with `setBackground()`, as [the documentation](https://github.com/PHPOffice/PHPPresentation/blob/develop/docs/usage/slides/introduction.md) shows.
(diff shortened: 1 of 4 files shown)


Progi1984:
Need a rebase

dkulyk:

Progi1984 (approved):

PR #909, opened 2026-08-20
A master slide is no longer born with a white background

dkulyk:
### Description

`SlideMaster::__construct()` sets a white `Background\Color` on every master it creates, and the PowerPoint2007 Writer turns it into `<p:bg>`. A converter paints that fill as a full-page path of its own, and the path lands outside any marked content -- which PDF/UA 7.1 refuses, "Content shall be marked as Artifact or tagged as real content". Measured with veraPDF on a six-slide deck: **6 failed checks, one per page**. A deck that draws its own backgrounds over it pays that price for a fill it never asked for.

Nothing is lost by leaving it out. The colour map of a master takes its background to `lt1`, and `lt1` is `FFFFFF` (`SlideMaster.php:63`, written by `PptTheme.php:102`), so the page renders exactly as before. I converted the same deck to PDF with and without the fill and compared the two renders pixel by pixel at 720x540: **no pixel differs in colour**. The only difference is in the alpha channel, 1259 bytes of 1 555 200, all of them on the outermost row and column of the page -- the antialiased edge of a fill that is no longer there.

A master that is given a background still writes one, and the Writer already returns early when there is none. This only stops the library from inventing one.

The ODPresentation Writer is untouched by the change: every place it reads a background takes it from a slide, never from the master, so saving the same presentation with and without the white fill produces byte-identical entries in the ODF package.

One existing expectation changed with it: `SlideMasterTest::testBase` asserted the white background, and now asserts that a fresh master carries none.

### Checklist:

- [x] My CI is :green_circle:
  > phpstan and php-cs-fixer are clean, and the suite passes: 810 tests, 7197 assertions.
  > `PowerPoint2007Test::testSlideMasterBackground` covers both sides -- no `p:bg` unless one is asked for, and `p:bg` written when it is. I checked it fails without the change, together with the updated `SlideMasterTest::testBase`.
  > `docs/usage/slides/introduction.md` gained a "Background" section: the API was not documented at all before.
  > Added an entry under 1.3.0, and a BC Breaks note, since a deck that relied on the white fill has to ask for it now.

### DIFF
--- a/docs/changes/1.3.0.md
+++ b/docs/changes/1.3.0.md
@@ -17,4 +17,7 @@
 - Fixed static analysis on PHP 8.4 and 8.5 (PHPStan 1.x could not resolve the symbols of PHPUnit 13) by [@yasumorishima](http://github.com/yasumorishima) in [#897](https://github.com/PHPOffice/PHPPresentation/pull/897)
 - ODPresentation Writer : Fixed the charts losing their data points, their background, their data labels and the colour of their series, and drawing a hidden legend by [@dkulyk](http://github.com/dkulyk) in [#901](https://github.com/PHPOffice/PHPPresentation/pull/901)
 - PowerPoint2007 Writer : Fixed the name of a slide being lost on save, and read it back by [@dkulyk](http://github.com/dkulyk) in [#902](https://github.com/PHPOffice/PHPPresentation/pull/902)
+- A master slide is no longer born with a white background nobody asked for by [@dkulyk](http://github.com/dkulyk) in [#909](https://github.com/PHPOffice/PHPPresentation/pull/909)
 
+## BC Breaks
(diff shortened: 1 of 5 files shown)


dkulyk:
Closing this: it is already on `master`. It was stacked under #912, and merging that one (a89f40fe3) carried every part of this PR in with it —

- `SlideMaster` no longer constructing a background — `src/PhpPresentation/Slide/SlideMaster.php`
- `testSlideMasterBackground` — `tests/…/Writer/PowerPoint2007Test.php:129`
- `self::assertNull($object->getBackground())` — `tests/…/Slide/SlideMasterTest.php:43`
- the `### Background` section in `docs/usage/slides/introduction.md:50`
- the changelog entry and the BC-break note in `docs/changes/1.3.0.md`

Nothing here is unmerged, so there is nothing left to review. Thanks!

PR #871, opened 2025-06-12
Powerpoint2007 reader: Added support for loading text run without properties

Wilkolicious:
### Description

Adds support for scenarios where a text run is stated without text run properties.

Also see: https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.drawing.runproperties?view=openxml-3.0.1

<details>
<summary>Example XML with text run without properties</summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<p:sld xmlns:a="http://schemas.openxmlformats.org/drawingml/2006/main"
    xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships"
    xmlns:p="http://schemas.openxmlformats.org/presentationml/2006/main"
    xmlns:m="http://schemas.openxmlformats.org/officeDocument/2006/math"
    xmlns:a14="http://schemas.microsoft.com/office/drawing/2010/main" showMasterSp="1"
    showMasterPhAnim="1">
    <p:cSld>
        <p:bg>
            <p:bgPr>
                <a:solidFill>
                    <a:srgbClr val="1893F8" />
                </a:solidFill>
            </p:bgPr>
        </p:bg>
        <p:spTree>
            <p:nvGrpSpPr>
                <p:cNvPr id="191" name="Google Shape;100;p2" />
                <p:cNvGrpSpPr />
                <p:nvPr />
            </p:nvGrpSpPr>
            <p:grpSpPr>
                <a:xfrm>
                    <a:off x="11797752" y="4618213" />
                    <a:ext cx="10123873" cy="2533542" />
                    <a:chOff x="0" y="0" />
                    <a:chExt cx="10123872" cy="2533541" />
                </a:xfrm>
            </p:grpSpPr>
            <p:sp>
                <p:nvSpPr>
                    <p:cNvPr id="185" name="Google Shape;101;p2" />
                    <p:cNvSpPr txBox="1" />
                    <p:nvPr />
                </p:nvSpPr>
                <p:spPr>
                    <a:xfrm>
                        <a:off x="0" y="0" />
                        <a:ext cx="1531523" cy="554881" />
                    </a:xfrm>
                    <a:prstGeom prst="rect">
                        <a:avLst />
                    </a:prstGeom>
                    <a:noFill />
                    <a:ln w="12700" cap="flat">
                        <a:noFill />
                        <a:miter lim="400000" />
                    </a:ln>
                    <a:effectLst />
                    <a:extLst>
[...]

PR #914, opened 2026-08-21
PowerPoint2007 Reader : Read a text run that was written without properties

dkulyk:
Supersedes #871.

### Description

The commit is @Wilkolicious's, from #871 — the diagnosis and the fix are his. This adds the tests, the guard his change makes necessary, and a rebase onto master.

**What breaks.** `a:rPr` is optional — `minOccurs="0"` in `CT_RegularTextRun` — and `loadParagraph()` wrapped both the creation of the run and its `setText()` in `if (is_object($oElementrPr))`. A run without properties was therefore skipped whole and its text lost, with no error.

**It is not hypothetical.** A deck exported from **Keynote** carries runs with no `a:rPr`, and it mixes them inside one paragraph:

```xml
<a:p>
  <a:pPr/>
  <a:r><a:rPr b="0"/><a:t>Quarterly </a:t></a:r>
  <a:r><a:t>review</a:t></a:r>
</a:p>
```

Read back, before and after:

| shape | master | this PR |
| --- | --- | --- |
| title | `'Quarterly '` | `'Quarterly review'` |
| subtitle | `''` | `'Prepared for the board'` |

The loss is ragged rather than total — a title breaks off mid-sentence — which is why it is easy to mistake for a layout problem. Google Slides writes `a:rPr` when a run has text, but omits it for empty placeholder runs: the same defect, harmless.

**One addition to #871.** Moving `setText()` out of the `is_object()` guard means it runs for every `a:r`, and `getElement()` returns `null` when there is no `a:t`. On master such a run is skipped whole; without a guard it would reach the dereference and the reader would be the one raising a warning about someone else's malformed file. `a:t` is required by the schema, so this is a guard against bad input, not a behaviour anyone should rely on.

**What this does not fix.** A run without `a:rPr` is not unstyled — it inherits, through `a:pPr/a:defRPr`, the shape's `a:lstStyle`, the placeholder's list style in the layout and master, and finally `p:defaultTextStyle`. The reader walks none of that chain, so the recovered text comes back as the library default (Calibri 10pt black) rather than what the file specifies. Separate gap, separate change; before this PR the text was not there at all.

### Checklist:

- [x] My CI is :green_circle:
  > 817 tests, 7226 assertions; phpstan clean.
  > `testTextRunWithoutProperties` and `testTextRunWithoutText`. Each writes a deck, removes `a:rPr` or `a:t` from the archive and reads it back, so neither carries a fixture. Both were checked by restoring the old code and watching them fail — the second needs an error handler to do it, since a PHP warning alone does not fail this suite.
  > Reader behaviour; no documented API changed.
  > Under *Bug fixes*, crediting both PRs.

### DIFF
--- a/docs/changes/1.3.0.md
+++ b/docs/changes/1.3.0.md
@@ -30,6 +30,7 @@
 - PowerPoint2007 Reader : Fixed the title of a chart axis, its rotation and the two fonts of the axis not being read back by [@dkulyk](http://github.com/dkulyk) in [#918](https://github.com/PHPOffice/PHPPresentation/pull/918)
 - PowerPoint2007 Reader : Fixed the minor tick mark of an axis being read into its major tick mark, and the outline of the value axis not being read at all, by [@dkulyk](http://github.com/dkulyk) in [#918](https://github.com/PHPOffice/PHPPresentation/pull/918)
 - Fixed the media part of an image whose path carries no extension being named with a trailing dot, and written under no content type, by [@dkulyk](http://github.com/dkulyk) in [#922](https://github.com/PHPOffice/PHPPresentation/pull/922)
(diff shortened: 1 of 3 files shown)


Progi1984 (approved):

Progi1984 (commented):
Need a rebase

Progi1984 (approved):

PR #916, opened 2026-08-22
PowerPoint2007 Reader : Read a slide number and a date with their styling

dkulyk:
### Description

Depends on #914 — it rewrites the body of the same loop. The first two commits are that PR (and @Wilkolicious's #871 under it); only the last one belongs to this review.

#### What is wrong

`a:fld` is a run whose text the application recomputes: the number of the slide it ended up on, the date the file is opened on. The Writer emits one for every `sldNum` or `dt` placeholder (`AbstractSlide.php:326`), but the Reader walked `(a:r|a:br)` only, so the element was skipped and everything inside it with it.

The shape survives the round trip — the placeholder is restored from `p:ph` further up, and the Writer re-derives the field from it — so a deck keeps its numbering. What it loses is how that number looks. Measured on a `sldNum` placeholder styled Georgia 18pt bold red:

| | before | after |
|---|---|---|
| placeholder read back | `sldNum` | `sldNum` |
| runs in the paragraph | **0** | 1 |
| `a:fld` on re-save | present, **no `a:rPr` at all** | full `a:rPr` |

`dt` behaves identically.

#### The change

`CT_TextField` holds the same `a:rPr` and `a:t` as `CT_RegularTextRun`, so the element is read by the branch that already reads a run — the XPath and the condition, nothing else. Both field types are covered at once, the kind being an attribute rather than a separate element.

#### One deliberate consequence

A shape that is *not* a placeholder but holds a field — `page <slidenum> of 12` in a plain text box — now reads its stand-in text as an ordinary run, where it used to be dropped. The text of the file is kept rather than lost, but a field is not expressible in the model, so re-saving that shape freezes it into dead text.

Both outcomes are wrong; I think keeping the text is the better of the two, and it is easy to reverse if you disagree. What actually closes it is a field element of its own next to `Run` and `BreakElement`, which is a larger change and a separate PR. For reference, that is how LibreOffice models it — `class TextField final : public TextRun`, and both its OOXML and its ODF exporters decide per run rather than per shape.

### Checklist:

- [x] My CI is :green_circle:
  > `PowerPoint2007Test::testFieldKeepsItsStyle`, over both field types; the fixture is the library's own output, so nothing binary is added. Each half of the fix was checked by breaking it — without `a:fld` in the XPath, and without it in the condition — and both cases fail with `actual size 0`.
  > Reader behaviour; no documented API changed. `docs/usage/slides/layout.md` already describes the fields, from #913.

### DIFF
--- a/docs/changes/1.3.0.md
+++ b/docs/changes/1.3.0.md
@@ -39,6 +39,7 @@
 - PowerPoint2007 Writer : Fixed `Axis::setFormatCode()` having no effect, the number format of an axis being taken from the source data instead by [@dkulyk](http://github.com/dkulyk) in [#903](https://github.com/PHPOffice/PHPPresentation/pull/903)
 - Fixed the fill of a table row never reaching the file, in both Writers, by [@dkulyk](http://github.com/dkulyk) in [#911](https://github.com/PHPOffice/PHPPresentation/pull/911)
 - PowerPoint2007 Writer : Fixed a cell losing its own right and bottom borders whenever it had a neighbour on that side by [@dkulyk](http://github.com/dkulyk) in [#906](https://github.com/PHPOffice/PHPPresentation/pull/906)
+- PowerPoint2007 Reader : Fixed a slide number and a date being read back without the font, size and colour they were written with by [@dkulyk](http://github.com/dkulyk) in [#916](https://github.com/PHPOffice/PHPPresentation/pull/916)
 
 ## BC Breaks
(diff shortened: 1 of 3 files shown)


Progi1984 (approved):

Progi1984:
Need a rebase

Progi1984 (approved):

dkulyk:
@Progi1984 Already rebased — `1d425c815` sits directly on `68ef0bf10` (#906), nothing behind.

PR #912, opened 2026-08-20
ODPresentation Writer : Keep the background of a master slide

dkulyk:
### Description

> Stacked on #909, which is where the white default disappears. Without it every ODF page would suddenly gain a white fill the deck never asked for. Please merge that one first; I will rebase this on master afterwards.

Every place the ODPresentation Writer reads a background takes it from a slide: `Styles.php` and `Pictures.php` and `MetaInfManifest.php` iterate `getAllSlides()`, and `Content.php` is handed one slide at a time by `writeStylePage()`. A background set on the master reached none of them, so it was dropped on save -- and the master is where PowerPoint usually keeps it.

`Styles.php` did write `<style:master-page … draw:style-name="sPres0">`, but `sPres0` was a hardcoded `style:family="presentation"` carrying `<style:graphic-properties draw:fill-color="#ffffff"/>`: the wrong family, the wrong property, and no `draw:fill` to switch it on. It painted nothing. Measured -- saving the same presentation with and without a background on the master produced byte-identical entries in the package.

That style is now what its name says: a drawing-page style built from the background of the master, as LibreOffice writes it.

```xml
<style:style style:name="sPres0" style:display-name="sPres0" style:family="drawing-page">
  <style:drawing-page-properties draw:fill="solid" draw:fill-color="#FFCC00"/>
</style:style>
```

**It has to be an automatic style.** I wrote it into `office:styles` first, next to where it used to be, and LibreOffice ignored it: the render was unchanged. The same style in `office:automatic-styles` -- where LibreOffice writes its own -- is drawn. Measured on the converted PDF:

| the master asks for | colour operators in the PDF | images in the PDF |
| --- | --- | --- |
| nothing | `0 0 0 rg`, `1 1 1 rg` | 0 |
| the colour `FFCC00` | **`1 0.8 0 rg`** | 0 |
| an image | `0 0 0 rg`, `1 1 1 rg` | **2** |

An image also has to reach the package and the manifest, so `Pictures.php` and `MetaInfManifest.php` now look at the master as well as at the slides, under the name `background_sPres0`. `writeBackgroundStyle()` takes that name as a `string` rather than an `int`, since the master is not numbered.

Along the way I checked a wrong guess so it does not come up in review: `presentation:background-visible="true"` on the style of the page changes nothing here. The section the style sits in was the whole of it.

### Checklist:

- [x] My CI is :green_circle:
  > phpstan and php-cs-fixer are clean, and the suite passes: 812 tests, 7212 assertions.
[...]

PR #917, opened 2026-08-22
ODPresentation Writer and Reader : Keep what a font says about itself

dkulyk:
### Description

Every `style:text-properties` this Writer builds was assembled by hand, and each one left something out.

| | italic | bold | underline | strikethrough |
|---|---|---|---|---|
| chart style (axis, axis title, legend, chart title) | kept | **lost** | **lost** | **lost** |
| chart style, series | **lost** | **lost** | **lost** | **lost** |
| text run | **lost** | kept | **lost** | **lost** |
| the same deck saved as PPTX | kept | kept | kept | kept |

Measured on one font set `b + i + u + strike`, saved both ways.

ODF splits these differently from OOXML: the underline (`style:text-underline-style` with a type, a width and a mode) and the strikethrough (`style:text-line-through-style` with a type) are one family for the whole run, while the family, the size, the weight and the italic are spelled once per script. The shared half now lives on the ODPresentation `AbstractDecoratorWriter` as `writeFontStates()`, with the table that turns each of the eighteen OOXML underlines into the style, type and width ODF names it with. The per-script half stays in the `switch` that already handled it, and the five chart blocks became one `writeTextProperties()` -- which is what closes the series block that was missing even the italic.

The Reader is the same defect read backwards: it took the family, the weight and the size, so a file this library had just written came back without its italic, its underline or its strikethrough. Round trip, before and after:

```
before   ODPresentation  bold=true  italic=false underline=none   strike=noStrike
after    ODPresentation  bold=true  italic=true  underline=sng    strike=sngStrike
         PowerPoint2007  bold=true  italic=true  underline=sng    strike=sngStrike
```

A pair ODF allows and OOXML has no word for is read as a plain single underline rather than dropped.

**Charts are untouched on the reading side** -- the PowerPoint2007 Reader takes only the name, the description and the geometry of a chart shape, and the ODPresentation Reader does not read charts at all. Both are separate holes.

### Checklist:

- [x] My CI is :green_circle:
  > 854 tests, 7513 assertions; phpstan clean; php-cs-fixer `0 of 276`.
  > `ObjectsChartTest::testFontStateOfEveryChartStyle` and `::testAxisFontUnderline` (all 17 mappings, each validated against the ODF 1.2 schema), `ContentTest::testRichTextRunFontState` (per script), `ODPresentationTest::testFontUnderlineSurvivesTheRoundTrip` and `::testFontStateSurvivesTheRoundTrip`. Checked by breaking each half -- dropping `fo:font-weight`, corrupting `WAVYHEAVY`, deleting `wave|double` from the Reader's table, and disabling the complex-script italic -- every mutation fails a test.
  > Writer and Reader behaviour; no documented API changed.

### DIFF
--- a/docs/changes/1.3.0.md
+++ b/docs/changes/1.3.0.md
@@ -22,6 +22,7 @@
 - ODPresentation Writer : Fixed the background of a master slide being dropped on save by [@dkulyk](http://github.com/dkulyk) in [#912](https://github.com/PHPOffice/PHPPresentation/pull/912)
 - PowerPoint2007 Reader : Fixed a shape that asked for no fill being read as having no fill at all, which crashes the ODPresentation Writer on save, by [@dkulyk](http://github.com/dkulyk) in [#915](https://github.com/PHPOffice/PHPPresentation/pull/915)
 - Fixed a file with no measurable size (an SVG, a video) warning on `setPath()`, and a fresh object being looked up in the hash table under a null key, by [@dkulyk](http://github.com/dkulyk) in [#920](https://github.com/PHPOffice/PHPPresentation/pull/920)
+- ODPresentation Writer and Reader : Fixed the italic, the bold, the underline and the strikethrough of a font being dropped on save and not read back, in a chart style and in a text run alike, by [@dkulyk](http://github.com/dkulyk) in [#917](https://github.com/PHPOffice/PHPPresentation/pull/917)
 
 ## BC Breaks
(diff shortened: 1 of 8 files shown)


Progi1984:
@dkulyk Can you rebase ?

dkulyk:
@Progi1984 There is nothing to rebase here — this one is already on `master`. It was the base of #923, and merging that PR (`8b74e7b8b`) squashed both commits in, this one included. Its changelog line went in with it and still credits this PR:

```
docs/changes/1.3.0.md:25  ...by @dkulyk in #917
```

Rebasing it now would not give an empty diff, it would give a harmful one: against current `master` this branch is 9 insertions and **54 deletions**, and the deletions are #923's work.

Your three `private` → `protected` suggestions are in as well — I applied them to this branch before #923 was rebased on top, so they rode along.

Same shape as #909 under #912. Safe to close whenever you like.

PR #918, opened 2026-08-22
PowerPoint2007 Reader : Read the text of a chart axis

dkulyk:
### Description

`c:catAx` and `c:valAx` were read for six things -- the orientation, whether the axis is deleted, the two tick marks, the label position and where it crosses -- and for none of the text. Round trip of a deck that styles its category axis:

| | before | after |
|---|---|---|
| `getTitle()` | `Axis Title` (the constructor default) | `Quarter` |
| `getTitleRotation()` | `0` | `45` |
| `getFont()` | Calibri 10pt black | Georgia 18pt bold italic double-underlined red |
| `getTickLabelFont()` | Calibri 10pt black | Verdana 7pt green |

Nothing else reads them either: the ODPresentation Reader does not read charts at all.

`a:defRPr` is the run properties of a run that does not exist yet, so it is read the way a run is read -- weight, italic, underline, strikethrough, size, baseline, colour and typeface -- into the two fonts an axis carries. The title text is every `a:t` of the rich text, and the rotation is the `rot` of its `a:bodyPr`. Both axes go through one `loadAxisText()`, because the two blocks that read them are copies of one another.

Which is how two more came to light, in the copied halves:

* **`c:minorTickMark` was assigned with `setMajorTickMark()`**, on both axes. A file written with `major=out, minor=in` came back as `major=in, minor=none`: the minor mark overwrote the major one and was itself lost.
* **The value axis passed `c:spPr/a:ln` to `loadStyleOutline()`**, which looks for an `a:ln` inside what it is given, so it never found one. An axis line written 3px green came back at the default 1px black. The category axis passes `c:spPr`, and that is the one that is right.

Both are one-line fixes in the blocks this PR was already rewriting, so they are in it; each has its own assertion.

Not touched, for the record: the writer's own slip at `PptCharts.php:2510`, where `strike` in the tick-label block reads `getFont()` while its seven neighbours read `getTickLabelFont()`. The Reader now faithfully reads back what that writes, which is how it showed up. It belongs with the larger question of which font an axis is styled with.

**One line of `phpmd.xml.dist`.** `Chart\Axis` in the signature of `loadAxisText()` is the class's 38th dependency, and the ceiling was 38, so the mess detector fails on the type hint alone. I raised it to 39, the way this file has been raised five times before (29 → 30 → 31 → 32 → 35 → 38, the last of them in #856). If you would rather not move it, the alternative is to drop the type declaration and leave only a `@param` -- say the word and I will.

### Checklist:

- [x] My CI is :green_circle:
  > 818 tests, 7247 assertions; phpstan clean; php-cs-fixer `0 of 276`.
  > `PowerPoint2007Test::testAxisTextIsReadBack` (title, rotation, both fonts, and that the two fonts stay separate) and `::testAxisTickMarksAndOutlineAreReadBack` (both axes). Checked by breaking each of the four fixes in turn -- suppressing the title, misspelling the tick-label XPath, restoring `setMajorTickMark()` and restoring `c:spPr/a:ln` -- every one fails an assertion.
  > Reader behaviour; no documented API changed.

### DIFF
--- a/docs/changes/1.3.0.md
+++ b/docs/changes/1.3.0.md
@@ -27,6 +27,8 @@
 - ODPresentation Writer and Reader : Fixed the italic, the bold, the underline and the strikethrough of a font being dropped on save and not read back, in a chart style and in a text run alike, by [@dkulyk](http://github.com/dkulyk) in [#917](https://github.com/PHPOffice/PHPPresentation/pull/917)
 - ODPresentation Writer : Fixed the tick labels of a chart axis being styled from `getFont()` rather than `getTickLabelFont()`, which is the font the PowerPoint2007 Writer uses for them, and fixed the strikethrough of those labels being read from the wrong font in the PowerPoint2007 Writer, by [@dkulyk](http://github.com/dkulyk) in [#923](https://github.com/PHPOffice/PHPPresentation/pull/923)
 - PowerPoint2007 Writer and Reader : Fixed the charset of a font being written in hexadecimal, which the schema forbids above 127 and the Reader read as decimal either way, by [@dkulyk](http://github.com/dkulyk) in [#919](https://github.com/PHPOffice/PHPPresentation/pull/919)
+- PowerPoint2007 Reader : Fixed the title of a chart axis, its rotation and the two fonts of the axis not being read back by [@dkulyk](http://github.com/dkulyk) in [#918](https://github.com/PHPOffice/PHPPresentation/pull/918)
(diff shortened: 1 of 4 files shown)


Progi1984 (approved):

Progi1984:
@dkulyk Need a rebase

dkulyk:
@Progi1984 Rebased on `955065ef9`.

Progi1984 (approved):

PR #919, opened 2026-08-22
PowerPoint2007 : Fixed the charset of a font being written in hexadecimal

dkulyk:
### Description

`CT_TextFont@charset` is an `xsd:byte`. The Writer spells it in hexadecimal (`AbstractSlide::writeRunStyles()`), the Reader reads it as decimal (`PowerPoint2007::loadParagraph()`), and the schema allows only the decimal form. Nobody agrees with anybody.

#### What is wrong

A charset never comes back as it went in:

| `setCharset()` | written | read back |
|---|---|---|
| 18 | `"12"` | 12 |
| 134 (GB2312) | `"86"` | 86 |
| 178 (Arabic) | `"b2"` | 0 |

The last row is the worse half: `b2` is not a value `xsd:byte` allows, so the file is invalid. `assertIsSchemaECMA376Valid()` would have said so, had any test set a charset above 127 — the one test there is sets 18 and asserts the `"12"`, which is how this survived.

#### What PowerPoint writes

Measured on a deck saved by PowerPoint, one run in Traditional Arabic and one in Wingdings:

```xml
<a:latin typeface="Traditional Arabic" panose="02020603050405020304" pitchFamily="18" charset="-78"/>
<a:cs    typeface="Traditional Arabic" panose="02020603050405020304" pitchFamily="18" charset="-78"/>
<a:latin typeface="Wingdings" pitchFamily="2" charset="2"/>
```

−78 is 178 as a signed byte; 2 needs no wrapping. Other decks in the same sample carry `charset="-122"` (134, GB2312) and `charset="-120"` (136, Big5). So the file spells the Windows charset ids above 127 as the negative byte they occupy, which is what the schema requires.

#### The change

The Writer writes the decimal byte — negative above 127 — and the Reader turns it back into the number the model holds. Nothing else moves; `Font::setCharset()` keeps taking the id the caller knows (178 for Arabic), which is what `Font::CHARSET_DEFAULT` and every Windows table mean by it.

### Checklist:

- [x] My CI is :green_circle:
  > `PptSlidesTest::testRichTextRunFontCharset()` now asserts the decimal it should always have asserted; `::testRichTextRunFontCharsetAboveAByte()` covers the byte that wraps, schema validation included; `PowerPoint2007Test::testFontCharsetSurvivesTheRoundTrip()` covers the pair over seven charsets (1, 0, 18, 134, 178, 204, 255). Each half was checked by breaking it: with the old Writer the wrapping test fails, with the old Reader four of the seven round trips fail.
  > No documented API changed — `setCharset()` takes the same numbers it always did.

### DIFF
--- a/docs/changes/1.3.0.md
+++ b/docs/changes/1.3.0.md
@@ -25,6 +25,7 @@
 - Fixed a file with no measurable size (an SVG, a video) warning on `setPath()`, and a fresh object being looked up in the hash table under a null key, by [@dkulyk](http://github.com/dkulyk) in [#920](https://github.com/PHPOffice/PHPPresentation/pull/920)
 - ODPresentation Writer and Reader : Fixed the italic, the bold, the underline and the strikethrough of a font being dropped on save and not read back, in a chart style and in a text run alike, by [@dkulyk](http://github.com/dkulyk) in [#917](https://github.com/PHPOffice/PHPPresentation/pull/917)
 - ODPresentation Writer : Fixed the tick labels of a chart axis being styled from `getFont()` rather than `getTickLabelFont()`, which is the font the PowerPoint2007 Writer uses for them, and fixed the strikethrough of those labels being read from the wrong font in the PowerPoint2007 Writer, by [@dkulyk](http://github.com/dkulyk) in [#923](https://github.com/PHPOffice/PHPPresentation/pull/923)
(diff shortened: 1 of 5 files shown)


Progi1984 (approved):

Progi1984:
@dkulyk Thank you for your contribution

Issue #819, opened 2024-09-05
imagecreatefromstring(): gd-png: libpng warning: iCCP: known incorrect sRGB profile


toto975:
For a pptm file, mime type application/vnd.openxmlformats-officedocument.presentationml.presentation

`IOFactory::load($fullPath)`
The error : imagecreatefromstring(): gd-png: libpng warning: iCCP: known incorrect sRGB profile

Here is the sample file (i renamed it .zip because .ppm is not supported in the attach files feature of github)
[iCCP.pptm.zip](https://github.com/user-attachments/files/16886328/iCCP.pptm.zip)

Progi1984:
https://stackoverflow.com/questions/70569197/imagecreatefrompng-gd-png-libpng-warning-interlace-handling-should-be-turne
https://github.com/dompdf/dompdf/pull/2750/files

Issue #741, opened 2023-05-24
getimagesizefromstring($imageFile) returns false causing exception when loading presentation


websuasive:
Exception thrown trying to load powerpoint presentation due getimagesizefromstring($imageFile) returning false, rather than a valid array.

(ErrorException(code: 0): Trying to access array offset on value of type bool at /var/www/html/vendor/phpoffice/phppresentation/src/PhpPresentation/Reader/PowerPoint2007.php:805)
[stacktrace]
#0 /var/www/html/vendor/phpoffice/phppresentation/src/PhpPresentation/Reader/PowerPoint2007.php(805): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError()
#1 /var/www/html/vendor/phpoffice/phppresentation/src/PhpPresentation/Reader/PowerPoint2007.php(1372): PhpOffice\\PhpPresentation\\Reader\\PowerPoint2007->loadShapeDrawing()
#2 /var/www/html/vendor/phpoffice/phppresentation/src/PhpPresentation/Reader/PowerPoint2007.php(506): PhpOffice\\PhpPresentation\\Reader\\PowerPoint2007->loadSlideShapes()
#3 /var/www/html/vendor/phpoffice/phppresentation/src/PhpPresentation/Reader/PowerPoint2007.php(396): PhpOffice\\PhpPresentation\\Reader\\PowerPoint2007->loadMasterSlide()
#4 /var/www/html/vendor/phpoffice/phppresentation/src/PhpPresentation/Reader/PowerPoint2007.php(353): PhpOffice\\PhpPresentation\\Reader\\PowerPoint2007->loadMasterSlides()
#5 /var/www/html/vendor/phpoffice/phppresentation/src/PhpPresentation/Reader/PowerPoint2007.php(169): PhpOffice\\PhpPresentation\\Reader\\PowerPoint2007->loadSlides()
#6 /var/www/html/vendor/phpoffice/phppresentation/src/PhpPresentation/Reader/PowerPoint2007.php(136): PhpOffice\\PhpPresentation\\Reader\\PowerPoint2007->loadFile()
#7 /var/www/html/app/Helpers/ParsePptx.php(35): PhpOffice\\PhpPresentation\\Reader\\PowerPoint2007->load()

Code with issue:

  $pathImage = implode('/', $pathImage);
  $imageFile = $this->oZip->getFromName($pathImage);
  if (!empty($imageFile)) {
[...]

PR #922, opened 2026-08-22
Shape\Drawing\File : Name an image file whose path does not name itself

### LINKED ISSUE #603: CreateDrawingShape cannot insert image without file extension
Hello,

took me a while to discover that when using CreateDrawingShape and then setting setPath to valid image without extension (i.e. ".jpg), the type of the image (JPEG) is not passed and fatal error occures: PHP Fatal error:  Uncaught Exception: Invalid parameters passed.
(in ...vendor/phpoffice/phppresentation/src/PhpPresentation/Writer/PowerPoint2007/ContentTypes.php:154)

dkulyk:
Closes #603.

#### What is wrong

`Shape\Drawing\File::getExtension()` derives the extension from the path alone:

```php
return pathinfo($this->getPath(), PATHINFO_EXTENSION);
```

A path is free to carry no extension — an upload stored under its hash, a file from `tempnam()`, an image served from a route — and then the answer is `''`. `getIndexedFilename()` builds the name of the media part out of it, so the part is called `logo1.`, ending in a bare dot. Both writers register content types by extension, so that part is written under none.

#### The change

The contents say what the name does not, and the same call already backs `getMimeType()` a few lines below:

```php
$image = getimagesizefromstring(CommonFile::fileGetContents($this->getPath()));

return is_array($image) ? (string) image_type_to_extension($image[2], false) : '';
```

It runs only when `pathinfo()` came back empty **and** the file exists, so a path that has an extension costs exactly what it cost before — no file is opened.

#### What it does not do

An extensionless SVG stays extensionless. `getimagesizefromstring()` reports nothing for it, and there is nothing else in the file for the decision to rest on. Neighbouring reports #741 and #819 are about the same family of images-loaded-from-content problems and are not touched here.

### Checklist:

- [x] My CI is :green_circle:
  > `FileTest::testPathWithoutAnExtension` — the repository's own PNG copied to a `tempnam()` path, so no binary fixture is added. Checked by putting the one-line `getExtension()` back: the test fails.
  > No documented API changed.

### DIFF
--- a/docs/changes/1.3.0.md
+++ b/docs/changes/1.3.0.md
@@ -29,6 +29,7 @@
 - PowerPoint2007 Writer and Reader : Fixed the charset of a font being written in hexadecimal, which the schema forbids above 127 and the Reader read as decimal either way, by [@dkulyk](http://github.com/dkulyk) in [#919](https://github.com/PHPOffice/PHPPresentation/pull/919)
 - PowerPoint2007 Reader : Fixed the title of a chart axis, its rotation and the two fonts of the axis not being read back by [@dkulyk](http://github.com/dkulyk) in [#918](https://github.com/PHPOffice/PHPPresentation/pull/918)
 - PowerPoint2007 Reader : Fixed the minor tick mark of an axis being read into its major tick mark, and the outline of the value axis not being read at all, by [@dkulyk](http://github.com/dkulyk) in [#918](https://github.com/PHPOffice/PHPPresentation/pull/918)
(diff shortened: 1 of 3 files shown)


Progi1984 (approved):

Progi1984:
@dkulyk Need a rebase

dkulyk:
@Progi1984 Rebased on `6e210d341`.

Progi1984 (approved):

PR #923, opened 2026-08-23
Chart axis : The tick labels are styled by the font meant for them

### LINKED ISSUE #717: Bar Chart - Change Axis Label Font
Hello,

Is it not possible to change the font and color of the labels on the x and y-axis? I can only figure out how to change the title. For example, I'd like the font/color of the axis labels to match the title in the image below, Arial and gray.

Thanks!

Screen Shot 2022-08-17 at 12 26 44 PM

### LINKED ISSUE #723: How to change font size of chart axis labels?
I am trying to set a different font size (and eventually change also the color) of X and Y axis labels in a chart.
My code is like:

```
$oShape = $currentSlide->createChartShape(); // ..etc. and chart looks good
$oShape->getPlotArea()->getAxisX()->getFont()->setSize(7); // it does nothing
```

setSize() doesn't set the new value and font size remains the same. No errors or warnings.

dkulyk:
> **Depends on #917.** That PR refactors the five chart text-property blocks onto a shared `writeTextProperties(Font $font)`, which this one hands a different font. The diff shown here includes its commit until it lands; review only the two one-line changes in `ODPresentation/ObjectsChart.php` and `PowerPoint2007/PptCharts.php`, and the tests.

### Description

Fixes #723 and #717.

`Shape\Chart\Axis` carries two fonts. `getTickLabelFont()` (`Axis.php:197`, added with the feature in `b72f395ac`, released in 1.1.0) styles the numbers along the axis; `getFont()` (`Axis.php:179`) styles the title the axis carries. The two Writers disagree about which is which:

| | tick labels | axis title |
| --- | --- | --- |
| PowerPoint2007 (`PptCharts.php:2495-2527` / `:2380-2461`) | `getTickLabelFont()` | `getFont()` |
| ODPresentation (`ObjectsChart.php` `writeAxisMainStyle()` / `writeAxisTitleStyle()`) | **`getFont()`** | `getFont()` |

`getTickLabelFont()` was not referenced anywhere in the ODPresentation Writer.

**Measured**, with `getAxisX()->getFont()->setSize(7)->setColor('FF0000')` and nothing else set:

```
chart1.xml            c:catAx/c:title/…/a:defRPr  sz="700"  srgbClr FF0000   ← the title, styled
                      c:catAx/c:txPr/…/a:defRPr   sz="1000" srgbClr 000000   ← the labels, untouched
Object 1/content.xml  styleAxisX                  fo:font-size="7pt" fo:color="#FF0000"   ← the labels
                      styleAxisXTitle             fo:font-size="7pt" fo:color="#FF0000"
```

That is the whole of #723 (and its duplicate #717): the reporter's `getAxisX()->getFont()->setSize(7)` changes nothing in a PPTX, and that is correct behaviour there — which is why the first question on the thread was which Writer they used. No single snippet worked for both formats.

### The change

ODF already keeps the two apart — `chart:axis` carries `chart:style-name="styleAxisX"` and the `chart:title` inside it carries `styleAxisXTitle` — so the fix is which font each style is built from. `writeAxisMainStyle()` takes `getTickLabelFont()`, `writeAxisTitleStyle()` keeps `getFont()`. One argument.

**A copy-paste slip on the other side, closed by the same change.** In the tick-label block of `writeAxis()`, `strike` alone read `getFont()` while its seven neighbours read `getTickLabelFont()`:

```php
2509: 'i',      $oAxis->getTickLabelFont()->isItalic()
2510: 'strike', $oAxis->getFont()->getStrikethrough()      // ← out of line
2511: 'sz',     $oAxis->getTickLabelFont()->getSize()
```

Introduced with the feature itself and untested until now, so a strikethrough asked for on the tick labels never arrived and one asked for on the axis title arrived twice.

### BC

A deck that sized or coloured its ODP tick labels through `getFont()` now gets the tick-label default (10pt Calibri black) until it also calls `setTickLabelFont()`. The recipe is two lines, and it is in the changelog under *BC Breaks*:

```php
$axis->getFont()->setSize(7);            // the axis title, as before
$axis->getTickLabelFont()->setSize(7);   // the tick labels, which used to follow the line above
```

The PowerPoint2007 side moves for nobody: `strike` was reading the wrong font, so any deck it affected was already getting something it did not ask for.

### Checklist:

- [x] My CI is :green_circle:
      `ObjectsChartTest::testAxisFont` now asserts the split — the tick-label font reaches `styleAxisX`, `getFont()` reaches `styleAxisXTitle`, and the two carry different values so neither can pass by accident. `testAxisFontUnderline` and `testFontStateOfEveryChartStyle` follow the same move; the latter now sets both fonts, since the axis no longer carries one font twice. On the PowerPoint2007 side `PptChartsTest::testAxisTickLabelFontStates` is new and covers the `strike` slip — it fails on master and passes here.
      The *Font* section of `docs/usage/shapes/chart.md` now says an axis has a font per half, with the two-line example.


src/PhpPresentation/Reader/ODPresentation.php
```diff
@@ -48,6 +48,31 @@
  */
 class ODPresentation implements ReaderInterface
 {
+    /**
+     * The reverse of what the Writer spells out: a style, a type and a width back into the single
+     * token OOXML names an underline with. `words` is not here -- ODF says it with a mode.
+     *
+     * @var array<string, string>
+     */
+    private const UNDERLINE_OOXML = [
```
Progi1984:
```suggestion
    protected const UNDERLINE_OOXML = [
```

src/PhpPresentation/Writer/ODPresentation/AbstractDecoratorWriter.php
```diff
@@ -20,11 +20,71 @@
...
 
 abstract class AbstractDecoratorWriter extends \PhpOffice\PhpPresentation\Writer\AbstractDecoratorWriter
 {
+    /**
+     * OOXML names an underline with a single token; ODF spells the same thing over a style, a type
+     * and a width, and has no name of its own for a few of the eighteen.
+     *
+     * @var array<string, array{0: string, 1: string, 2: null|string}>
+     */
+    private const UNDERLINE_ODF = [
```
Progi1984:
```suggestion
    protected const UNDERLINE_ODF = [
```

src/PhpPresentation/Writer/ODPresentation/AbstractDecoratorWriter.php
```diff
@@ -20,11 +20,71 @@
...
+        Font::UNDERLINE_WAVY => ['wave', 'single', null],
+        Font::UNDERLINE_WAVYDOUBLE => ['wave', 'double', null],
+        Font::UNDERLINE_WAVYHEAVY => ['wave', 'single', 'bold'],
+        Font::UNDERLINE_WORDS => ['solid', 'single', null],
+    ];
+
+    /**
+     * @var array<string, string>
+     */
+    private const STRIKETHROUGH_ODF = [
```
Progi1984:
```suggestion
    protected const STRIKETHROUGH_ODF = [
```

dkulyk:
Thanks — all three applied, but in #917 rather than here, since that is the PR those lines belong to; they only show up in this diff because this one is stacked on it. GitHub will not mark the suggestions as accepted for that reason.

`UNDERLINE_OOXML`, `UNDERLINE_ODF` and `STRIKETHROUGH_ODF` are `protected` now. This branch is rebased on top, so the change is in both diffs.

Progi1984 (approved):

PR #924, opened 2026-08-24
Shape\Chart and RichText : Replace a font set to null with a fresh one

dkulyk:
Every font setter in the model is declared `?Font $font = null` and stores what it is given. Nothing writes a font back and no writer guards the read, so one line of user code takes both writers down.

### What fails today

Measured on `3848ad3c9`, one line of user code each, both writers:

| call | PowerPoint2007 | ODPresentation |
| --- | --- | --- |
| `$axis->setFont(null)` | `isBold() on null` | `TypeError` on `writeTextProperties()` |
| `$axis->setTickLabelFont(null)` | `isBold() on null` | `TypeError` on `writeTextProperties()` |
| `$legend->setFont(null)` | `getHashCode() on null` | `getHashCode() on null` |
| `$title->setFont(null)` | `getHashCode() on null` | `getHashCode() on null` |
| `$series->setFont(null)` | `isBold() on null` | `TypeError` on `writeTextProperties()` |
| `$run->setFont(null)` | `TypeError` on `getFont()` | `getHashCode() on null` |
| `$paragraph->setFont(null)` | writes | `getHashCode() on null` |

`Run` is the sharpest case: `getFont(): Font` promises a font the setter is allowed to take away, so the failure is a `TypeError` raised by the getter's own return type.

### The fix

A setter given `null` puts a fresh `Font` in place — which is what every constructor already does — and the getters lose their `?`. That makes the declared `?Font` mean *reset*, the only sense in which it can work. The two guards in the Reader that tested those getters for null go with it, since they can no longer be false.

`BreakElement::getFont()` and `TextElement::getFont()` keep returning `?Font`: those elements carry no font by design, and `Run` narrows the return type when it overrides.

### Tests

The unit tests for five of these classes asserted the defect (`assertNull($object->getFont())` after `setFont()`); they now assert the font that replaces it. Two writer-level tests save a deck with every one of the seven fonts nulled, through each writer, and check the output still validates against its schema.

Without the `src/` change, those tests give 3 errors and 6 failures; with it the suite is green — 878 tests, 7631 assertions, under `--fail-on-warning --fail-on-notice --fail-on-deprecation --fail-on-phpunit-deprecation`.

### BC

`Axis::getFont()`, `Axis::getTickLabelFont()`, `Title::getFont()`, `Legend::getFont()`, `Series::getFont()` and `Paragraph::getFont()` return `Font` rather than `?Font`. Safe for callers; a subclass that overrode one with a nullable return type widens it. Recorded in the changelog.

### Checklist:

- [x] My CI is :green_circle:
  > The unit tests of five of the six classes asserted the defect outright (`assertNull($object->getFont())` after `setFont()`); they now assert the font that replaces it. `RunTest::testFont` gains the no-argument call it never made. Two writer-level tests, `PptChartsTest::testFontSetBackToNull` and `ObjectsChartTest::testFontSetBackToNull`, save a deck with every one of the seven fonts nulled. Checked from both sides: with the `src/` change reverted, those give 3 errors and 6 failures.
  > No documented API changed. The six getters narrowed from `?Font` to `Font`, which the changelog records under BC Breaks.
  > A bug-fix line, and a BC Breaks entry for the narrowed return types.


Progi1984 (approved):
Need a rebase

dkulyk:

Progi1984 (approved):

PR #925, opened 2026-08-24
ODPresentation Writer and Reader : Keep the columns of a text shape

dkulyk:
`RichText::setColumns()` and `setColumnSpacing()` are listed in [the documentation](https://github.com/PHPOffice/PHPPresentation/blob/master/docs/usage/shapes/richtext.md) as properties of a rich-text shape, with no note that either is writer-specific. In the whole of `src/` they had exactly two readers, and both were the PowerPoint2007 Writer:

```
Writer/PowerPoint2007/AbstractSlide.php:307   if ($shape->getColumns() != 1) {
Writer/PowerPoint2007/AbstractSlide.php:308       writeAttribute('numCol', $shape->getColumns());
Writer/PowerPoint2007/AbstractSlide.php:309       writeAttribute('spcCol', pixelsToEmu($shape->getColumnSpacing()));
```

The ODPresentation Writer never asked for them. One shape, `setColumns(3)->setColumnSpacing(20)`, saved through both writers before this change:

```
PPTX  <a:bodyPr … numCol="3" spcCol="190500">
ODP   no style:columns anywhere; the frame's graphic style is
      <style:style style:name="gr1" style:family="graphic" style:parent-style-name="standard">
        <style:graphic-properties style:mirror="none" draw:fill="none" draw:fill-color="#000000"
          draw:stroke="none" fo:wrap-option="wrap"/>
      </style:style>
```

The setter accepts, the model keeps, one writer honours it and the other drops it without a word.

### This is not a format limitation

ODF 1.2 has the feature: `style:columns` carries `fo:column-count` (required) and `fo:column-gap` (optional), and it is a legal child of `style:graphic-properties` — which is exactly the element `writeTxtStyle()` already opens for the frame's own `gr{shapeId}` style. It is an element rather than an attribute, so it is written after the attributes of that block. A shape left with the single column it starts with writes nothing, which is the same guard the PowerPoint2007 Writer uses.

```xml
<style:graphic-properties style:mirror="none" … fo:wrap-option="wrap">
  <style:columns fo:column-count="3" fo:column-gap="0.529cm"/>
</style:graphic-properties>
```

### The Reader is the other half

`loadShapeRichText()` read the frame's geometry and its paragraphs and never looked at `draw:style-name` at all — so the graphic style of a text frame was parsed into `arrayStyles` and then ignored, while `loadShapeDrawing()` next to it has been reading its own style for shadow and fill all along. It now takes the two values back out of that map.

That makes the round trip exact: 20 pixels of gap are written as `0.529cm` and read back as 20, since `centimetersToPixels()` rounds.

### Tests

- `ContentTest::testRichTextColumns` — a single-column shape writes no `style:columns`; a three-column one writes the count and the gap. Both states validate against the ODF 1.2 schema.
- `ODPresentationTest::testTextColumns` — write, read, compare, with a second shape that never asked for columns to pin the default.

Suite is green — 880 tests, 7649 assertions, under `--fail-on-warning --fail-on-notice --fail-on-deprecation --fail-on-phpunit-deprecation`. PHPStan and PHP-CS-Fixer both clean.

No BC: a deck that never called `setColumns()` writes exactly what it wrote before.

### Checklist:

- [x] My CI is :green_circle:
  > `ContentTest::testRichTextColumns` — a single-column shape writes no `style:columns`, a three-column one writes the count and the gap; both states are checked against the ODF 1.2 schema. `ODPresentationTest::testTextColumns` — write, read, compare, with a second shape that never asked for columns to pin the default.
  > `docs/usage/shapes/richtext.md`: `columns` had a line in the property list and no section of its own, and `columnSpacing` said "for a paragraph" where it belongs to the shape.

### DIFF
--- a/docs/changes/1.3.0.md
+++ b/docs/changes/1.3.0.md
@@ -42,6 +42,7 @@
 - PowerPoint2007 Writer : Fixed a cell losing its own right and bottom borders whenever it had a neighbour on that side by [@dkulyk](http://github.com/dkulyk) in [#906](https://github.com/PHPOffice/PHPPresentation/pull/906)
 - PowerPoint2007 Reader : Fixed a slide number and a date being read back without the font, size and colour they were written with by [@dkulyk](http://github.com/dkulyk) in [#916](https://github.com/PHPOffice/PHPPresentation/pull/916)
 - Fixed a font set to `null` -- which every font setter accepts -- crashing both Writers on save, by [@dkulyk](http://github.com/dkulyk) in [#924](https://github.com/PHPOffice/PHPPresentation/pull/924)
+- ODPresentation Writer and Reader : Fixed the columns of a text shape being dropped on save and never read back, by [@dkulyk](http://github.com/dkulyk) in [#925](https://github.com/PHPOffice/PHPPresentation/pull/925)
 
 ## BC Breaks
 - `\PhpOffice\PhpPresentation\Slide\SlideMaster` is constructed without a background. A deck that relied on the white fill it used to write sets one with `setBackground()`, as [the documentation](https://github.com/PHPOffice/PHPPresentation/blob/develop/docs/usage/slides/introduction.md) shows.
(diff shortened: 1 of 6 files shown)


Progi1984 (commented):
Need rebase

dkulyk:

PR #926, opened 2026-08-24
Order the columns of a text shape, instead of always left to right

dkulyk:
### Description

`rtlCol` on `a:bodyPr` is written as `'0'` at three sites in the PowerPoint2007 Writer, with no setter, no documentation and no test. Only one of the three sits on a shape that can have columns at all — `writeShapeText()`, sixteen lines above the `numCol` it belongs with. The other two are a table cell and an autoshape, neither of which has a `setColumns()`, so they stay `'0'`.

ECMA-376 calls the attribute *Columns Right-To-Left* (`dml-main.xsd:8299`): it orders the columns of a multi-column text body and says nothing about the direction of the text, which `a:pPr@rtl` already carries correctly. So a deck whose text reads right to left had its columns laid out left to right, and no way to say otherwise.

**The setter is nullable on purpose.** It follows `setAutoShrinkVertical()` in the same class:

```php
$richText->setColumnsRTL(true);   // right to left, whatever the text does
$richText->setColumnsRTL(false);  // left to right, whatever the text does
$richText->setColumnsRTL();       // null, the default: take the order from the paragraphs
```

Deriving the order from the paragraphs and stopping there would have been smaller, and it would have made a legal document unreachable from the model: **left-to-right text in right-to-left columns**. PPTX spells it `rtlCol="1"` alongside `rtl="0"` on each `a:pPr`; ODF spells it `style:writing-mode="rl-tb"` on the frame while every paragraph style carries its own `lr-tb`. The two are independent in both schemas, so `null` derives and an explicit value overrides. The derivation needs the paragraphs to be non-empty and all right to left — "every element satisfies it" is vacuously true on an empty array, which would have put `rtlCol="1"` on an empty shape.

`hasColumnsRTL()` gives back what was set, `null` included; `isColumnsRTL()` gives the order the writers use.

**Both writers, not just one.** ODF carries the same thing as `style:writing-mode` on the frame's `style:graphic-properties` — valid there per the schema (`OpenDocument-v1.2-os-schema.rng:17007`, through `common-writing-mode-attlist`), and written by LibreOffice itself on every frame it saves. It cannot leak into the text: the neighbouring `writeTxtStyle()` code writes `style:writing-mode` on every paragraph style unconditionally, so each paragraph always states its own direction. A PPTX-only setter would be an API one writer honours and the other silently drops, which is the defect #925 just fixed for `setColumns()` itself.

Both writers state the order for every shape, as the PowerPoint2007 Writer already did for `rtlCol`, so a value that was never set comes back said outright after a round trip. Nothing that was already correct changes: a deck with left-to-right paragraphs writes `rtlCol="0"` and `style:writing-mode="lr-tb"` exactly as before.

**What LibreOffice does with it, measured, since it is worth being straight about:** nothing. It drops `rtlCol` on a PPTX round trip (the attribute is absent from its output entirely) and resets a frame's `style:writing-mode` to `lr-tb` on an ODP round trip. It models column order in neither format, so it can neither confirm nor contradict the change; the references here are the two schemas and PowerPoint's own reading of `rtlCol`.

> [!IMPORTANT]
> **Stacked on #925 — please do not merge this one first.** Its base is `feature/odp-text-columns`, because the `style:columns` element it orders is what #925 adds. Against `master` alone the ODP half of this PR would write a writing mode for columns that do not exist yet. Once #925 is in I will rebase this onto `master`; the diff below is then this commit alone.

Fixes # (issue)

### Checklist:

- [x] My CI is :green_circle:
  > `RichTextTest::testColumnsRTL` and `testColumnsRTLFromParagraphs` for the tri-state and the derivation, including the empty-shape case. `PptSlidesTest::testParagraphColumnsRTL` and `ContentTest::testRichTextColumnsRTL` for the two writers, in all three states, schema-checked; the ODP one also pins that the paragraph keeps its own writing mode when the frame disagrees with it. `PowerPoint2007Test::testColumnsRTL` and `ODPresentationTest::testTextColumnsRTL` write, read and compare, each with a second shape that never asked for anything.
  > `docs/usage/shapes/richtext.md` — the *Columns* section #925 added gains the order, and `columnsRTL` joins the property list.

### DIFF
--- a/docs/changes/1.3.0.md
+++ b/docs/changes/1.3.0.md
@@ -43,6 +43,7 @@
 - PowerPoint2007 Reader : Fixed a slide number and a date being read back without the font, size and colour they were written with by [@dkulyk](http://github.com/dkulyk) in [#916](https://github.com/PHPOffice/PHPPresentation/pull/916)
 - Fixed a font set to `null` -- which every font setter accepts -- crashing both Writers on save, by [@dkulyk](http://github.com/dkulyk) in [#924](https://github.com/PHPOffice/PHPPresentation/pull/924)
 - ODPresentation Writer and Reader : Fixed the columns of a text shape being dropped on save and never read back, by [@dkulyk](http://github.com/dkulyk) in [#925](https://github.com/PHPOffice/PHPPresentation/pull/925)
+- Fixed the columns of a text shape always being written left to right, and added `RichText::setColumnsRTL()` to say the order outright, by [@dkulyk](http://github.com/dkulyk) in [#926](https://github.com/PHPOffice/PHPPresentation/pull/926)
 
 ## BC Breaks
 - `\PhpOffice\PhpPresentation\Slide\SlideMaster` is constructed without a background. A deck that relied on the white fill it used to write sets one with `setBackground()`, as [the documentation](https://github.com/PHPOffice/PHPPresentation/blob/develop/docs/usage/slides/introduction.md) shows.
--- a/docs/usage/shapes/richtext.md
+++ b/docs/usage/shapes/richtext.md
@@ -19,6 +19,7 @@ Below are the properties that you can set for a rich text shape.
 - `upright`
 - `vertical`
 - `columns` see *Columns*
+- `columnsRTL` see *Columns*
 - `bottomInset` in pixels
 - `leftInset` in pixels
 - `rightInset` in pixels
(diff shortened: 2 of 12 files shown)


dkulyk:
Left as a draft on purpose, and it is not waiting on review feedback.

This branch is stacked on #925: its base is `feature/odp-text-columns`, so the two commits above are that PR's and this one's. GitHub can only open it against `master`, which is why both show here.

The reason is not just tidiness. The ODP half of this change writes `style:writing-mode` on the frame's `style:graphic-properties` to order the columns that `style:columns` declares — and `style:columns` is what #925 adds. On `master` alone it would be a writing mode for columns that the writer does not yet emit.

The moment #925 is merged I will rebase this onto `master` and take it out of draft; the diff then becomes the single commit `Order the columns of a text shape, instead of always left to right`. Nothing else is outstanding — CI is green and it is ready to read now if you would rather review it before then.

dkulyk:
#925 is in, so this is rebased onto `master` and out of draft. The diff is now this PR's own work — `rtlCol` on the PowerPoint2007 side, `style:writing-mode` on the ODPresentation side, and the reader for each.

Progi1984 (approved):

PR #927, opened 2026-08-24
ODPresentation Reader : Read back the alignment of a paragraph

dkulyk:
### Description

`loadStyle()` parses `style:paragraph-properties` into an `Alignment` and files it in `arrayStyles` under `alignment`. In the whole of the Reader that key has exactly one reader — `readListItem()` — and the object it reads there is a *different* one, built from `text:list-style` for the indent and the margins. So for an ordinary paragraph the alignment was parsed and then dropped: `readParagraph()` takes `spacingAfter`, `spacingBefore`, `lineSpacingMode` and `lineSpacing` out of the same style entry and walks straight past the alignment sitting next to them.

Three things had to be right for the round trip. Two were not.

**`fo:text-align` was stored raw.** `setHorizontal($nodeParagraphProps->getAttribute('fo:text-align'))` puts the ODF spelling — `left`, `right`, `center`, `justify` — into a model whose constants are `l`, `r`, `ctr` and `just`. So even on the one path that did use this object, the value could never match what the Writer switches on. It is mapped now, over the four alignments this library writes. (`start` and `end` are deliberately not mapped: their meaning depends on the direction of the text, and LibreOffice writes `left` — checked on a file it saved — so nothing here produces them.)

**`style:writing-mode` was read as left to right in both arms of its switch.** `lr-tb`/`tb-lr`/`lr` and `rl-tb`/`tb-rl`/`rl` alike called `setIsRTL(false)`; the second is a typo for `true`, present since be48bb383 added right-to-left support in July 2021. On its own it changes nothing observable, which is exactly why it survived four years: with no paragraph ever reading the alignment, the value it corrects was never asked for by anything.

**And `readParagraph()` now applies the alignment it already had in hand.**

Measured before the change, on a shape written by this library: `style:writing-mode="rl-tb"` goes into `content.xml`, and `isRTL()` comes back `false`. After it, all four horizontal alignments and the direction survive a save and a load.

Independent of #925 and #926 — different region of the file, `git merge-tree` clean against both, and the full suite passes on each.

Fixes # (issue)

### Checklist:

- [x] My CI is :green_circle:
  > `ODPresentationTest::testHorizontalAlignmentSurvivesTheRoundTrip` over the four alignments, and `testTextDirectionSurvivesTheRoundTrip` with a second shape reading the other way to pin the arm next to it. Checked from both sides: with the `src/` change reverted they give 4 failures — `left` passes either way, since it is what an unread alignment defaults to.
  > No documented API changed; `alignment` is already listed as a paragraph property, and this is the Reader catching up with it.

### DIFF
--- a/docs/changes/1.3.0.md
+++ b/docs/changes/1.3.0.md
@@ -33,6 +33,7 @@
 - PowerPoint2007 Reader : Fixed a text run written without `a:rPr` being dropped, which loses part of the text of a Keynote export, by [@Wilkolicious](http://github.com/Wilkolicious) in [#871](https://github.com/PHPOffice/PHPPresentation/pull/871) and [@dkulyk](http://github.com/dkulyk) in [#914](https://github.com/PHPOffice/PHPPresentation/pull/914)
 - `Style\Color` : Fixed a colour written as a plain RGB being given an alpha it never asked for by [@dkulyk](http://github.com/dkulyk) in [#910](https://github.com/PHPOffice/PHPPresentation/pull/910)
 - PowerPoint2007 Reader : Fixed a slide size written with an explicit `type="custom"` crashing the reader instead of being read with its own dimensions by [@SAY-5](http://github.com/SAY-5) in [#929](https://github.com/PHPOffice/PHPPresentation/pull/929)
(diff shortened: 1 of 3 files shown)


Progi1984 (approved):

Progi1984:
@dkulyk Need a rebase

dkulyk:

Progi1984 (approved):

PR #928, opened 2026-08-24
Replace a shadow set to null with a fresh one, rather than storing the null

dkulyk:
### Description

`AbstractShape::setShadow()` takes `?Shadow` and stores whatever it is given, so `getShadow()` returns `?Shadow`. Every writer reads it without asking whether it is there, and every one of them declares a non-nullable `Shadow` parameter. Measured on a shape that asked for its shadow to be taken away — `$shape->setShadow(null)`, which the signature invites:

| written as | result |
| --- | --- |
| HTML, rich text | OK — this writer reads no shadow from a text body |
| PowerPoint2007 | `TypeError: AbstractSlide::writeShadow(): Argument #2` |
| ODPresentation | `TypeError: Content::writeStylePartShadow(): Argument #2` |
| HTML, image | `TypeError: HTML::getStyleShadow(): Argument #1` |
| HTML, video | `TypeError: HTML::getStyleShadow(): Argument #1` |

Five crash sites across three writers — `AbstractSlide.php:273` and `:1534`, `Content.php:1011` and `:1142`, `HTML.php:232` and `:301` — all from one setter.

**Why the setter and not the call sites.** Guarding each one is five guards for a state the model should not hold in the first place. A shape is *born* with `new Shadow()` in the constructor; a shape asked to give its shadow up has no business becoming a shape that has none, because `Shadow` already carries `isVisible()` for exactly that. So the setter replaces a null with a fresh `Shadow`, `getShadow()` narrows to `Shadow`, and all five sites are correct without touching any of them.

This is the same treatment `setFont()` got in #924, for the same reason: a setter that accepts null is a promise the writers cannot keep.

The `isset($this->shadow)` guard in `__clone()` goes with it — the property is always set now, so the clone is unconditional, as it already was for `$border`. (PHPStan catches this one: *Property `$shadow` in isset() is not nullable*.)

**Credit where it is due.** @phili67 found this in #887 and fixed one of the five sites from inside the HTML writer. That PR guards `writeImage()` but not `writeVideo()` twelve lines further down, and its second hunk adds `is_null($style)` inside `getStyleShadow(Shadow $style)` — unreachable under a non-nullable parameter, which is what its red `PHP Static Analysis` and `PHP CS Fixer` checks are about. The changelog credits both.

Fixes # (issue)

### Checklist:

- [x] My CI is :green_circle:
  > `AbstractShapeTest::testShadow` asserted the defect outright (`assertNull($object->getShadow())` after a bare `setShadow()`); it now asserts the shadow that replaces it, and adds the explicit `setShadow(null)` the old test never made. Three writer-level tests save a deck whose shadows were nulled — `HTMLTest::testSaveShadowSetBackToNull` (image and video), `PptSlidesTest::testShadowSetBackToNull` and `ContentTest::testShadowSetBackToNull` (text body and autoshape, schema-checked). Checked from both sides: with the `src/` change reverted, those give 3 errors and 1 failure.
  > No documented API changed; the return type narrowed, which the changelog records.
  > A bug-fix line crediting both authors, plus a BC Breaks entry for the narrowed return type.


Progi1984 (commented):
Could you check in each Writer that any shadow content is written ?

Progi1984 (approved):

PR #932, opened 2026-08-25
Honour a border that was given no colour, instead of crashing on it

dkulyk:
### Description

`Border::setColor()` accepts a `null` and stores it, but every reader of the colour wanted a `Color`. Setting a line style as well as the null was enough to take down both Writers, and the hash on its own:

```php
$shape->getBorder()->setLineStyle(Border::LINE_SINGLE)->setColor(null);
```

| read by | before |
| --- | --- |
| PowerPoint2007 | `TypeError: writeColor(): Argument #2 ($color) must be of type Color, null given` |
| ODPresentation | `Error: Call to a member function getRGB() on null` |
| `Border::getHashCode()` | `Error: Call to a member function getHashCode() on null` |

It has gone unreported because a shape is born at `Border::LINE_NONE`, and `writeBorder()` returns before it reaches the colour in that case.

**The null keeps its place in the model, rather than being replaced at the setter.** That is deliberately not what #924 did for a null `Font`, #928 proposes for a null `Shadow`, or #931 did for a null `Fill`. Those three were fixed at the setter because a null font, fill or shadow is a promise no Writer can keep — there is no way to write "no font named" into either format. A null line *colour* is different: both formats say it.

- OOXML — `EG_LineFillProperties` is `minOccurs="0"` in `CT_LineProperties`, so an `a:ln` carrying `w`, `cmpd` and `a:prstDash` but no fill of its own is valid, and means the line takes its colour from the theme.
- ODF — `svg:stroke-color` is optional, and `fo:border` is the CSS2 shorthand whose colour is the optional third part.

Replacing the null with a fresh black would invent a decision the caller did not make, and would *override* the theme rather than defer to it. So the five places that read the colour are guarded instead, and no signature moves.

| site | now |
| --- | --- |
| `AbstractDecoratorWriter::writeBorder()` | the `a:solidFill` is skipped |
| `Content.php`, graphic style | no `svg:stroke-color` |
| `Content.php`, `Line` shape | no `svg:stroke-color` |
| `Content.php::getBorderValue()` | `fo:border` is written as a width and a style |
| `Border::getHashCode()` | the colour contributes nothing |

The hash is not cosmetic: it fails with no Writer involved, and it is what the ODP table path reaches *first*, because the four borders of a cell are compared by hash before any of them is written.

What comes out:

```xml
<a:ln w="38100" cap="flat" cmpd="sng" algn="ctr">
  <a:prstDash val="solid"/><a:round/>…
</a:ln>
```

```xml
fo:border-bottom="0.57pt solid"
```

Both validate against the schemas the test suite checks.

There is no BC break: `setColor(null)` crashes today, so there is no behaviour to preserve, and `getColor(): ?Color` is unchanged. One asymmetry is documented rather than fixed — the PowerPoint2007 Reader reads a border written without a colour as the black a `Border` starts with, because nothing is left in the file to say the colour was never named.

Fixes # (no issue)

### Checklist:

- [x] My CI is :green_circle:
> Locally: `phpunit` 942 tests / 7906 assertions with `--fail-on-warning --fail-on-notice --fail-on-deprecation --fail-on-phpunit-deprecation`, `phpstan` clean, `php-cs-fixer` 0 of 276.
> Four: `BorderTest::testHashCodeWithoutAColor`, `PptSlidesTest::testDrawingShapeBorderWithoutAColor`, `ContentTest::testRichTextBorderWithoutAColor` and `ContentTest::testTableCellBorderWithoutAColor`. The three Writer tests assert the output against the ECMA-376 and OpenDocument 1.2 schemas.
> `docs/usage/styles.md`, the Border section: what a null colour means, an example, and the Reader asymmetry.
> Under Bug fixes in `docs/changes/1.3.0.md`. Nothing under BC Breaks — no signature moves.


Progi1984 (commented):
Need a rebase

dkulyk:
@Progi1984 Rebased onto `master`, thanks. The head is now `ee89d5500`, on top of `9f0d66017`, with nothing behind it.

Progi1984 (approved):

PR #935, opened 2026-08-26
ODPresentation Writer : Share one automatic style between everything that writes it

### LINKED ISSUE #888: Missing/wrong styles in ODP when using slide notes
In an ODP file with slide notes, some shapes are missing their styles or have wrong styles applied. Here's why:

When ````content.xml```` gets written, the Writer first outputs all the styles. For each shape, ````$this->shapeId```` is incremented and a corresponding shape style "grXX" is written. However, this pass does not count note shapes!

In a second pass, the actual document body is written, shape by shape. For each shape (including note shapes), ````$this->shapeId```` is incremented and style "grXX" (if applicable) is referenced. Since the note shapes are counted here, starting from the first slide with notes, the shapeIds will be out of sync with those used in the first pass. This leads to wrong or missing styles. If no style is found, LO will apply an (ugly) default style to the shape.

I guess the solution would be to include slide note shapes in the first pass as well. If they're RichText, probably a style should be written for them anyway?

dkulyk:
### Description

The ODPresentation Writer names a paragraph style `P_` plus `Paragraph::getHashCode()`, and a text
style `T_` plus `Run::getHashCode()`. **`getHashCode()` is the identity of the content, and a style
is the identity of the formatting.** `Cell::getHashCode()` folds in the hash of each of its
paragraphs, `Paragraph::getHashCode()` folds in the hash of each of its runs — so two paragraphs
formatted alike but holding different text hash differently, which is every paragraph of a real
document.

The result is one definition per paragraph and one per run, shared with nothing. On a slide holding
a 10×10 table with distinct text and twenty rich-text shapes:

| | `master` | this PR |
| --- | --- | --- |
| `style:family="paragraph"` definitions | 20 | **1** |
| `style:family="text"` definitions | 120 | **1** |
| references to them | 140 | 140 |
| `content.xml` | 96 314 bytes | **54 805 bytes** |

This keys the style on the XML it writes, rendered to a scratch `XMLWriter`, so two styles are the
same exactly when they write the same thing. The name becomes a prefix and a counter — `P1`, `T1` —
which are the prefixes LibreOffice generates (`xmloff/source/text/txtparae.cxx:1470-1481`), so a
file this writer produces is named the way one LibreOffice produces is.

A generated name follows from the order the styles were added, so the pass that writes the content
cannot recompute it the way it recomputed a hash. It reads it from a map keyed by the object wearing
the style, which is what `XMLTableExport` does in LibreOffice
(`maCellStyleMap`, `xmloff/source/table/XMLTableExport.cxx:312` and `:445`).

**And it fixes a slide note referencing three styles nobody wrote.** The pass that collects styles
walks the shapes of each slide and never the shapes of its note, while the pass that writes the
content walks both. On `master` a note names a paragraph style, a text style *and* a graphic style
that are not in the file, and its shape puts the counter the two passes share out of step:

```
ref stylePage0   -> defined      ref gr2          -> *** DANGLING ***
ref gr1          -> defined      ref P_dbead106…  -> *** DANGLING ***
ref P_dc1c4dcc…  -> defined      ref T_5db0da2d…  -> *** DANGLING ***
ref T_478fc3fe…  -> defined
```

Collecting the note's shapes with the rest fixes all three. This is not optional here — a pooled
name cannot be invented for an object that was never collected, so the gap had to be closed.

**A group inside a group is the same omission, and it was already there.** The collecting pass walks
a group but stops at one nested in it, while the pass that writes the content recurses. On `master`
that leaves every shape inside a nested group with three dangling references and the shape counter
out of step from there on:

```
$outer = new Group();
$inner = new Group();
$inner->createRichTextShape()->createTextRun('Nested');
$outer->addShape($inner);
$slide->addShape($outer);
```

```
master:   10 styles defined, 3 dangling: gr6, P_5b4e587c…, T_fd69ea6e…
this PR:   4 styles defined, 0 dangling
```

Keyed on the object rather than on a counter, a name that used to be merely wrong becomes no name at
all — the writer raises a `TypeError` — so this had to be closed here too, for the same reason the
note did.

The bullet family (`L_`) and the table families (`gr{n}r{n}c{n}`) are deliberately left alone; they
are separate steps.

### Built on #934

#934 has merged, so this is rebased onto `master` and out of draft — one commit, ready to read.

It needed #934 first: every ODPresentation test that asserted on a generated style name spelled that name out or recomputed it, so renaming `P_<hash>` to `P1` broke 54 assertions. With #934 in place resolving each name from the reference the document carries, **not one test needed changing here** — the 152 ODPresentation writer tests pass untouched.

### Checklist:

- [x] My CI is :green_circle:
  > 943 tests, 8188 assertions, PHPStan level 4 and PHP-CS-Fixer all clean locally.
  > `testStyleIsSharedByEverythingThatWritesIt()`, `testStyleIsNotSharedBetweenRunsThatDiffer()`,
  > `testSlideNoteNamesStylesThatAreWritten()` and `testNestedGroupNamesStylesThatAreWritten()`. All
  > but the second fail on the base commit.
  > Nothing to update: no public API changes. What changes is the name a generated style is given
  > and how many of them are written, neither of which the documentation describes.
  > One line under Bug fixes. No BC Breaks line for the generated names — they are automatic styles,
  > internal to the file. Two `protected` properties on `Writer\ODPresentation\Content`,
  > `$arrStyleParagraph` and `$arrStyleTextFont`, are gone, replaced by the pool; say the word if you
  > would rather that were called out.


dkulyk:
#934 has merged, so this is rebased onto `master` (`ebafa02ad`, on `29b6662b3`) and out of draft. One commit now, and no test in it needed changing. Ready for review.

Progi1984 (approved):

PR #936, opened 2026-08-26
ODPresentation Writer : Share the graphic and the drawing-page styles too

dkulyk:
### Description

[#935](https://github.com/PHPOffice/PHPPresentation/pull/935) put the paragraph and the text
families in a style pool. The other two families the ODPresentation Writer generates were left
naming a style after the *position* of the thing wearing it — `gr` plus the shape counter for
`style:family="graphic"`, `stylePage` plus the slide index for `style:family="drawing-page"`. A
positional name cannot be shared, so a document writes one definition per shape and one per slide
however alike they are.

Ten default slides, each holding eight identically formatted text shapes and a line:

| | #935 | this PR |
| --- | --- | --- |
| `style:family="graphic"` definitions | 90 | **2** |
| `style:family="drawing-page"` definitions | 10 | **1** |
| definitions, all families | 102 | **5** |
| `content.xml` | 57 632 bytes | **32 225 bytes** |

The names are `gr1`, `dp1` — the prefixes LibreOffice uses for the two families.

**Nothing about the file changes but the names and the count.** Resolving every style reference in
document order to the full body it points at, on two documents built to exercise the writer:

| | references | identical | different | definitions | bytes |
| --- | --- | --- | --- | --- | --- |
| uniform | 260 | **260** | 0 | 102 → **5** | 57 632 → **32 225** |
| varied | 85 | **85** | 0 | 38 → **34** | 22 440 → **21 886** |

The varied document is the control: five slides of five shapes that differ in fill, border,
alignment and size. Its thirty `graphic` definitions stay thirty — the pool shares a drawing-page
and nothing else, because there is nothing else alike to share.

A side effect worth naming: the two passes no longer have to count in step to agree on a graphic or
a drawing-page name. That lockstep is what `$this->shapeId` was reset and re-incremented for; it
stays only for the chart objects, which are genuinely addressed by their index.

The table families keep their positional names on purpose. Renaming them wants a round trip to hold
it to, and there is none — `Reader\ODPresentation` does not read a `table:table` at all, so a table
this library writes reads back as nothing. That reader is the missing half of
[#911](https://github.com/PHPOffice/PHPPresentation/pull/911)/[#915](https://github.com/PHPOffice/PHPPresentation/pull/915)
and belongs before the table families move, not after.

### Built on #935

#935 has merged, so this is rebased onto `master` and out of draft — two commits, ready to read.

It needed #935 first: it calls `shareAutomaticStyle()` and `getAutomaticStyleName()`, which #935 introduces.

`Content.php` has taken ten changes since, and the rebases carried all of them in: the `svg:stroke-color` guards of #932 (twice), its optional colour in `getBorderValue()`, the bullet-type guard of #933, the line shadow of #937, the list numbering of #945, the hyperlink of #944, the header row of #942, the pattern fill of #957, the field of #959 and the part reference of #948. Two sit inside bodies this PR moves into a closure; the rest are in shape writers it does not touch. The pool keys on exactly the XML a style writes, so a border that names no colour simply hashes to a style of its own.

The reference-resolution table above was re-measured on the current `master`, after all ten: every one of the 345 references still resolves to a byte-identical body.

### Checklist:

- [x] My CI is :green_circle:
  > 1115 tests, 9145 assertions, PHPStan level 4, PHPMD and PHP-CS-Fixer all clean locally.
  > No new test was needed and none had to change: with #934 in place every assertion about a
  > generated style resolves the name from the reference the document carries, so renaming
  > `gr{shapeId}` to `gr1` and `stylePage{n}` to `dp1` is invisible to the suite. #934 also brings
  > the `drawing-page` family under that rule and gives it the round trip it had no test for.
  > Nothing to update: no public API changes. What changes is the name a generated style is given
  > and how many are written, neither of which the documentation describes.
  > One line under Bug fixes.


dkulyk:
Left as a draft on purpose — not waiting on review.

It is stacked on #935 (itself stacked on #934), which is why three commits show in the diff. On `master` alone it does not build against itself: it calls `shareAutomaticStyle()` and `getAutomaticStyleName()`, which #935 introduces.

It will be rebased and undrafted the moment #935 merges. Please do read it now.

dkulyk:
#935 has merged, so this is rebased onto `master` (`e85b136f6`, on `c6986a597`) and out of draft. One commit now.

The rebase carried in the four changes `Content.php` has taken since this branch forked — #932's `svg:stroke-color` guards and its optional colour in `getBorderValue()`, #933's bullet-type guard and #937's line shadow — two of them inside bodies this PR moves into a closure. 982 tests, PHPStan and PHP-CS-Fixer green. Ready for review.

src/PhpPresentation/Writer/ODPresentation/Content.php
```diff
@@ -169,44 +171,17 @@ protected function writeContent(): string
         $incSlide = 0;
         foreach ($this->getPresentation()->getAllSlides() as $pSlide) {
             // Slides
-            $this->writeStyleSlide($objWriter, $pSlide, $incSlide);
+            $this->addSlideStyle($pSlide, $incSlide);
 
             // Images
```
Progi1984:
```suggestion
            // Shapes
```

dkulyk: Taken, thank you. `addShapeStyles()` walks the whole shape collection, so the comment was reading one kind of shape onto all of them.

There is a second `// Images` further down, at the `draw:page` loop in `writeSlides()`, over the same collection and just as stale -- both say `// Shapes` now.

Progi1984 (commented):
Please rebase it :)

dkulyk:
@Progi1984 Rebased onto `master`, on top of #949.

Progi1984 (approved):

PR #938, opened 2026-08-26
PowerPoint2007 Writer : Write the hyperlink a shape was given, whatever the shape is

dkulyk:
### Description

`PptSlides::writeSlide()` (`:145`) walks the shape collection and asks `hasHyperlink()` on
`AbstractShape`, so it writes the hyperlink relationship for **every** kind of shape. Only
`writeShapeText()` (`:221`) and `writeShapePic()` (`:1413`) ever wrote the `a:hlinkClick` that points
at it.

So a link set on anything else was dropped — and the slide still shipped the relationship. Measured
on `3ad8ec8b0`, one autoshape and one text shape, each given a URL:

```
hlinkClick count in slide1.xml: 1
hyperlink relationships:        2
```

`rId2` names `https://example.com/autoshape` and nothing in `slide1.xml` refers to `rId2`. That is
the part that makes this a defect rather than a missing feature: the file carries a relationship
nobody points at.

Measured by putting a hyperlink on one shape of each kind and looking for it in the output:

| shape | `a:hlinkClick` | relationship |
| --- | --- | --- |
| RichText | yes | yes |
| Drawing | yes | yes |
| Line | no → **yes** | yes |
| AutoShape | no → **yes** | yes |
| Table | no → **yes** | yes |
| Chart | no → **yes** | yes |
| Group | no → **yes** | yes |
| a text *run* | yes | yes |

The last row is the one that always worked, and is probably why nobody has filed this: the
documented way to link from a deck is `createTextRun()->getHyperlink()`, and that path is complete.
`AbstractShape::getHyperlink()` is the one that mostly was not.

Five cells, five one-line calls into a `p:cNvPr` each writer already opens: `p:nvCxnSpPr` for a
line, `p:nvSpPr` for an autoshape, `p:nvGraphicFramePr` for a table and a chart, `p:nvGrpSpPr` for a
group. `writeHyperlink()` already returns early when there is no hyperlink, so none of the five
needs a guard around it.

**`a:hlinkClick` goes before `a:extLst`**, which is the order `CT_NonVisualDrawingProps` wants and
what `writeShapeText()` already does. It only shows on a shape that is also marked decorative,
because that is what puts an `a:extLst` in the same `p:cNvPr` —
`testAutoShapeHyperlinkPrecedesTheDecorativeExtension()` is that shape, and moving the call one line
down turns it red on the ECMA-376 schema.

**What is not here.** The ODPresentation Writer has the same gap in six cells and spells a shape
link as `office:event-listeners/presentation:event-listener` — `Content.php:510` already writes
exactly that, for a picture and nothing else. That is a separate PR. The HTML Writer has no
shape-level link at all, only the run-level `<a href>`; it is not planned.

### Checklist:

- [x] My CI is :green_circle:
  > 959 tests, 7942 assertions, PHPStan level 4 and PHP-CS-Fixer all clean locally.
  > `PptSlidesTest::testLineHyperlink()`, `::testAutoShapeHyperlink()`, `::testTableHyperlink()`,
  > `::testChartHyperlink()`, `::testGroupHyperlink()` and
  > `::testAutoShapeHyperlinkPrecedesTheDecorativeExtension()`. All six fail on the base commit and
  > all six validate against the ECMA-376 schema. Three of them also assert that the `r:id` written
  > is the one the slide's `.rels` carries, so the relationship is no longer dangling.
  > Nothing to update: no public API changes. `AbstractShape::getHyperlink()` is already there on
  > every shape; this makes the Writer read it.
  > One line under Bug fixes.


Progi1984 (approved):

Progi1984:
Need a rebase

dkulyk:
@Progi1984 Rebased onto `master`, thanks. The head is now `c085f115f`, on `00a9f62a9`.

Progi1984 (approved):

PR #940, opened 2026-08-26
Address a chart style by the name the document references, not by one the test spells out

dkulyk:
### Description

[#934](https://github.com/PHPOffice/PHPPresentation/pull/934) closed this hole for `ContentTest` and `StylesTest` and named `ObjectsChartTest` as the part left over. This is that part.

Every assertion about a generated chart style went straight to the **definition**, by a name the test spells out:

```php
$element = '/office:document-content/office:automatic-styles/style:style[@style:name=\'styleAxisX\']/style:chart-properties';
```

Nothing looked at the `chart:style-name` that names it. **So a style nothing points at passes.**

Measured by replacing every `chart:style-name` the Writer writes with a literal `BROKEN`, leaving each definition correctly named — a chart in which not one reference resolves:

| | `master` | this PR |
| --- | --- | --- |
| `ObjectsChartTest` | **58 of 61 pass** | 4 of 61 pass |
| `Reader/ODPresentationTest` | 51 of 51 pass | 51 of 51 pass |

`assertIsSchemaOpenDocumentValid('1.2')` passes on that file as well: the OpenDocument relax-ng does not check that a style reference resolves.

The three tests that did notice are the mirror image of the problem — they pin the *reference* to a literal and never look at the definition. The two halves never met.

### The fix

The name is read from the reference, the way a consumer reads it, using the `getZipXmlAttributeValue()` helper #934 put on `PhpPresentationTestCase`. One resolver plus nine thin wrappers, each naming where the reference lives:

| wrapper | reference |
| --- | --- |
| `getChartStyleXPath()` | `chart:chart/@chart:style-name` |
| `getTitleStyleXPath()` | `chart:chart/chart:title` |
| `getLegendStyleXPath()` | `chart:chart/chart:legend` |
| `getPlotAreaStyleXPath()` | `chart:chart/chart:plot-area` |
| `getAxisStyleXPath($dimension)` | `chart:axis[@chart:dimension='x']` |
| `getAxisTitleStyleXPath($dimension)` | `chart:axis[…]/chart:title` |
| `getGridlinesStyleXPath($dimension, $class)` | `chart:axis[…]/chart:grid[@chart:class='major']` |
| `getSeriesStyleXPath($series)` | `chart:plot-area/chart:series[1]` |
| `getDataPointStyleXPath($point, $series)` | `chart:series[1]/chart:data-point[1]` |

48 spelled-out names over 11 replaced, plus the five-name loop in `testChartFontStates()` and the gridlines data provider — which keeps its literal assertion on the reference and resolves the definition from it, so both properties hold.

Verified in both directions: breaking a reference fails 57 tests, and renaming a definition alone (`styleChart` → `styleChartRENAMED`, reference untouched) fails the two tests that read it.

`ObjectsChartTest` goes from 874 to **1020 assertions** — resolving a name asserts that the reference node exists and that it carries the attribute.

**One trap worth naming.** Nine tests built their XPath as the first statement of the test, before the chart shape existed. That is harmless with a literal name and impossible once the helper has to write the presentation to resolve one. Those statements moved to after the chart is set up; nothing else about those tests changed.

**No production code is touched**, and no chart style is renamed. This only makes the suite hold the Writer to the references it already writes — the net that has to exist before the names can ever move.

While measuring, three definitions turned out to have no test of their own at all: `styleWall`, `styleFloor` and `styleAxisYTitle`. Left alone here; this PR converts what exists rather than adding coverage.

### Checklist:

- [x] My CI is :green_circle:
> 1066 tests, 9009 assertions, PHPStan level 4, PHPMD and PHP-CS-Fixer all clean locally.
> The change is test-only. What it adds is the coverage the numbers above measure: 58 of 61 tests passing on a fully dangling document becomes 4 of 61.
> Nothing to update: no production code and no public API is touched.
- [ ] I have updated the [changelog](https://github.com/PHPOffice/PHPPresentation/blob/develop/docs/changes/1.1.0.md)
> Nothing here is visible to a user of the library, the same as #934.

### DIFF
--- a/tests/PhpPresentation/Tests/Writer/ODPresentation/ObjectsChartTest.php
+++ b/tests/PhpPresentation/Tests/Writer/ODPresentation/ObjectsChartTest.php
@@ -50,6 +50,8 @@ use PHPUnit\Framework\Attributes\DataProvider;
  */
 class ObjectsChartTest extends PhpPresentationTestCase
 {
+    private const CHART = '/office:document-content/office:body/office:chart/chart:chart';
+
     protected $writerName = 'ODPresentation';
 
     /**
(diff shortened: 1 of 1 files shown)


Progi1984 (commented):
Please rebase

dkulyk:
@Progi1984 Rebased onto `master`, green at 26/26.

Progi1984 (approved):

PR #952, opened 2026-08-31
PowerPoint2007 Writer and Reader : Write the pattern a pattern fill names, and read it back

### LINKED ISSUE #951: PowerPoint2007 Writer : a pattern fill is written without its pattern, and no constant could name one
A pattern fill comes out of the PowerPoint2007 Writer with no pattern in it, and there is no
`FILL_PATTERN_*` constant that could have put one there.

## The attribute is never written

`a:pattFill` names the pattern it draws with `prst`. `writePatternFill()`
(`Writer/PowerPoint2007/AbstractDecoratorWriter.php`) opens the element, writes `a:fgClr` and
`a:bgClr`, and closes it. The attribute is written nowhere.

```php
$shape->getFill()
    ->setFillType(Fill::FILL_PATTERN_LIGHTGRID)
    ->setStartColor(new Color('FF4472C4'))
    ->setEndColor(new Color('FFFFFFFF'));
```

comes out as

```xml
<a:pattFill>
  <a:fgClr><a:srgbClr val="4472C4"/></a:fgClr>
  <a:bgClr><a:srgbClr val="FFFFFF"/></a:bgClr>
</a:pattFill>
```

PowerPoint opens the file without complaint and paints the shape flat. Nothing warns, because
`CT_PatternFillProperties` has `prst` as `use="optional"` — the element is schema-valid with no
pattern in it. Every pattern fill this library has ever written is one of these.

## The constants name the wrong vocabulary

The seventeen `FILL_PATTERN_*` values in `Style\Fill` are the `ST_PatternType` of SpreadsheetML:

```php
public const FILL_PATTERN_DARKDOWN = 'darkDown';
public const FILL_PATTERN_LIGHTGRID = 'lightGrid';
public const FILL_PATTERN_MEDIUMGRAY = 'mediumGray';
```

They came over from PHPExcel. DrawingML names a preset pattern by `ST_PresetPatternVal`, which has
54 values — `pct5`, `ltDnDiag`, `lgGrid`, `solidDmnd`, `diagBrick`, `wdDnDiag`, … — and shares
**not one** of them with `ST_PatternType`. So even had `prst` been written, all seventeen constants
would have written an attribute the schema refuses. There is no way to say `wdDnDiag` at all.

For comparison, a deck saved by PowerPoint itself carries exactly what is missing here:

```xml
<a:pattFill prst="wdDnDiag">
```

## The Reader has no branch for it

`PowerPoint2007::loadStyleFill()` handles `a:solidFill`, `a:gradFill`, `a:blipFill` and
`a:noFill`. A pattern fill falls through and is read as no fill at all, so a deck that PowerPoint
wrote loses its patterns on load as well as on save.

## Suggested fix

- Give `Fill` the 54 `ST_PresetPatternVal` presets, named after the value each writes.
- Keep the seventeen old names as deprecated aliases of the preset each was asking for
  (`DARKDOWN` → `dkDnDiag`, `LIGHTGRID` → `smGrid`, `MEDIUMGRAY` → `pct50`, …). Their values change,
  which is a BC break, but the value they had could never be written; nobody has a file relying on
  it.
- Write `prst` when the fill type is one of the presets, and leave it out otherwise, so a fill type
  the library does not recognise comes out exactly as it does today rather than as an invalid
  attribute.
- Read `a:pattFill` back, with its two colours.

I have this working and will open a PR against it.

dkulyk:
### Description

A pattern fill has never had a pattern in it. `a:pattFill` names the pattern it draws with `prst`,
and `writePatternFill()` wrote the two colours and not the attribute, so every pattern fill this
library has written came out flat. It is schema-valid — `CT_PatternFillProperties` has `prst` as
`use="optional"` — which is why nothing ever complained.

The seventeen `FILL_PATTERN_*` constants could not have written a pattern either. Their values
were the `ST_PatternType` of SpreadsheetML (`darkDown`, `lightGrid`, `mediumGray`), inherited from
PHPExcel, and that vocabulary shares not one value with the `ST_PresetPatternVal` DrawingML names a
preset pattern by. All seventeen were invalid there.

`Fill` now carries the 54 presets the format defines, named after the value each writes
(`FILL_PATTERN_WDDNDIAG`, `FILL_PATTERN_LGGRID`, `FILL_PATTERN_SOLIDDMND`, …), and keeps the
seventeen old names as deprecated aliases of the preset each was asking for. `Fill::PATTERN_TYPES`
lists them, which is how the Writer tells a pattern fill from a fill type it does not know: a type
outside the list is written as it was before, rather than as an attribute the schema would refuse.

The Reader had no branch for `a:pattFill`, so a pattern fill written by PowerPoint was read as no
fill. It reads one now, with its foreground and background colours, and goes on ignoring an element
that names no preset — there is no fill type to read that one back as.

The BC break is the value of the seventeen aliases. It is noted in the changelog; the value they
carried could never be written into a file, so no deck depends on it.

Fixes #951

### Checklist:

- [x] My CI is :green_circle:


dkulyk:
The red job is not this branch. `PHP Static Analysis (7.4)` dies inside PHPStan itself:

```
PHP Fatal error: Uncaught Error: Call to undefined function str_contains() in
phar://.../phpstan.phar/src/Turbo/TurboExtensionSelector.php:113
```

PHPStan 2.2.11 went out at 10:34 UTC today, declares `php: ^7.4|^8.0`, and calls a PHP 8.0 function before it analyses anything. `composer.lock` is not committed, so every job takes the newest release and every pull request opened since then fails that one check. Upstream: [phpstan/phpstan#15137](https://github.com/phpstan/phpstan/issues/15137).

#953 excludes the single release and leaves the constraint at `^2.2`. Merging it and re-running this branch turns the check green; the other 25 already are.

Progi1984 (commented):
Need a rebase After the merge of the another one

Progi1984 (approved):

dkulyk:
@Progi1984 Already rebased onto master after #955 landed, and green at 26/26.

The conflict was in `PowerPoint2007Test.php` alone: #955 and this branch both appended a test right after `testSeriesDataPoints`, so the two bodies were interleaved. I took the master side whole and re-added `testSeriesDataPointPatternFillIsReadBack` as its own method.

#936 is rebased too, and #957 (the ODPresentation half of this) sits on top of this branch as a draft -- I will rebase it onto master and mark it ready once this one is merged.

Progi1984 (approved):

PR #955, opened 2026-08-31
PowerPoint2007 Reader : Read the eight kinds of chart it could not, and the data a series holds

### LINKED ISSUE #954: PowerPoint2007 Reader : eight of the nine chart types are unreadable, and the ninth comes back with no data
A chart written by this library cannot be read back by it. Two separate reasons, both in
`Reader\PowerPoint2007`.

## Eight of the nine plot types are not recognised

`loadShapeChart()` looks for one element:

```php
if ($oElement = $xmlReader->getElement('/c:chartSpace/c:chart/c:plotArea/c:barChart')) {
```

`Writer\PowerPoint2007\PptCharts` writes nine: `c:areaChart`, `c:barChart`, `c:bar3DChart`,
`c:doughnutChart`, `c:lineChart`, `c:pieChart`, `c:pie3DChart`, `c:radarChart`, `c:scatterChart`.
For the other eight nothing calls `setType()`, so the plot area is left without one and the next
thing to touch it throws:

```php
$presentation = new PhpPresentation();
$doughnut = new Doughnut();
$doughnut->addSeries(new Series('Downloads', ['Jan' => '1', 'Feb' => '5']));
$presentation->getActiveSlide()->createChartShape()->getPlotArea()->setType($doughnut);

$file = tempnam(sys_get_temp_dir(), 'pp');
(new PowerPoint2007Writer($presentation))->save($file);
(new PowerPoint2007Reader())->load($file);
```

```
PhpOffice\PhpPresentation\Exception\UndefinedChartTypeException
  in src/PhpPresentation/Shape/Chart/PlotArea.php:96
```

## The series comes back with no data, including for the bar chart

Even the one chart that is read loses everything it holds. The Reader looks for

```php
$xmlReader->getElement('c:tx/c:strRef/c:strCache/c:pt/c:v', $elementSerie)   // title
$xmlReader->getElement('c:cat/c:strRef/c:strCache', $elementSerie)           // categories
$xmlReader->getElement('c:val/c:numRef/c:numCache', $elementSerie)           // values
```

which is how a chart that points at a worksheet keeps a copy of what it last read there. This
library writes charts *without* a worksheet, and `writeMultipleValuesOrReference()` then writes the
literal form -- `c:strLit`, `c:numLit` -- with the title as a bare `c:tx/c:v`. None of the three
xpaths match, so the round trip above gives a `Bar` whose series has an empty title and
`getValues() === []`.

A scatter chart is a third case again: it calls its two axes `c:xVal` and `c:yVal`, which the
Reader looks for nowhere.

## Suggested fix

- Look the plot type up by element name, over the nine the Writer knows.
- Read the settings that belong to a kind of chart by what that kind can be told -- the direction,
  grouping, gap and overlap of a bar; the hole size and first slice angle of a doughnut; the
  explosion of a pie; the smoothing of a line -- and read the `c:ser` loop once for all nine, since
  it never depended on the kind of chart it sat in.
- Accept the reference form and the literal form for the title, the categories and the values, and
  accept `c:xVal`/`c:yVal` alongside `c:cat`/`c:val`.

I have this working and will open a PR against it.

dkulyk:
### Description

A chart written by this library could not be read back by it, for two separate reasons.

**Eight of the nine plot types were not recognised.** `loadShapeChart()` looked for `c:barChart`
and nothing else; the Writer writes nine elements. For the other eight nothing called `setType()`,
so the plot area was left without one and the next thing to touch it threw
`UndefinedChartTypeException`.

`c:plotArea` holds exactly one of those elements, and the name of it is the whole of what says
which chart this is -- the series, the categories and the labels underneath are the same either
way. So the type is looked up by element name, the settings that belong to a kind of chart are read
by what that kind can be told (the direction, grouping, gap and overlap of a bar; the hole size and
first slice angle of a doughnut; the explosion of a pie; the smoothing of a line), and the `c:ser`
loop, which never had anything to do with the kind of chart it sat in, is read once for all nine.

**The series came back empty, including for the bar chart that was read.** The Reader looked for
`c:strRef/c:strCache` and `c:numRef/c:numCache`, which is how a chart that points at a worksheet
keeps a copy of what it last read there. This library writes charts without a worksheet, and those
hold their points themselves in `c:strLit` and `c:numLit`, with the title as a bare `c:tx/c:v`.
None of the three xpaths matched, so a round trip gave back a series with no title and no values at
all. Both forms are read now, and so are the `c:xVal` and `c:yVal` a scatter chart calls its two
axes.

`CouplingBetweenObjects` goes from 43 to 47: the nine kinds of chart are nine classes, and the
comment above the property says so alongside the reasons already there.

Fixes #954

### Checklist:

- [x] My CI is :green_circle:

The documentation never said which charts could be read, so there is no page that is now wrong.

Note that `PHP Static Analysis (7.4)` will be red until #953 lands: PHPStan 2.2.11 cannot start on
PHP 7.4 ([phpstan/phpstan#15137](https://github.com/phpstan/phpstan/issues/15137)).


Progi1984:
@dkulyk Can you rebase ?

dkulyk:
@Progi1984 Rebased onto master, now that #953 has landed. 26/26 green.

I rebased #952 onto the same commit as well, since it was red on that PHPStan job for the same reason. It is 26/26 green too.

Progi1984 (approved):

PR #959, opened 2026-09-01
Shape\RichText : Let a field sit inside a sentence, not only be a whole shape

### LINKED ISSUE #958: A field can only be a whole shape, never a word inside a sentence
A field is a run whose text the application recomputes -- the number of the slide it sits on, the
number of slides there are, the date the presentation is opened on. In OOXML it is `a:fld`, and it
stands beside `a:r` and `a:br` inside `a:p`:

```xml
<a:p>
  <a:r><a:t>page </a:t></a:r>
  <a:fld id="{A4004C23-…}" type="slidenum"><a:rPr b="1"/><a:t>&lt;nr.&gt;</a:t></a:fld>
  <a:r><a:t> of </a:t></a:r>
  <a:fld id="{5B1194E2-…}" type="slidecount"><a:t>12</a:t></a:fld>
</a:p>
```

This library cannot express that. The PowerPoint2007 Writer emits `a:fld` from the **shape**, not
from the text: `AbstractSlide.php:323-357` writes one when the whole shape is a placeholder of type
`PH_TYPE_SLIDENUM` or `PH_TYPE_DATETIME`, taking the run properties from the first element of the
first paragraph. So a whole shape can be a slide number, but a field cannot sit inside a sentence
-- "page 7 of 12" is inexpressible. There is no field class in `Shape/RichText/` at all: the writer
never needed one.

The Reader half of this was fixed in #916, which reads `a:fld` the way it reads `a:r` and so keeps
the styling. But it gives back a plain `Run`, so the *kind* of field is dropped: read a mixed
paragraph and save it, and every field in it comes out as dead text.

## What the reference implementation does

LibreOffice decides per run rather than per shape. `oox/source/export/drawingml.cxx:3160-3237`
(`DrawingML::WriteRun`):

```cpp
if( bWriteField )
    mpFS->startElementNS( XML_a, XML_fld, XML_id, sUUID.getStr(), XML_type, sFieldValue );
else
    mpFS->startElementNS( XML_a, XML_r );
```

Everything after that -- the run properties, `a:t` -- is the shared path. Nothing in either filter
consults the placeholder. The class is `oox/inc/drawingml/textfield.hxx:31`,
`class TextField final : public TextRun`, adding a type and a uuid.

Round-tripping a shape whose paragraph is `page <slidenum> of <slidecount> at <datetime3>`:

| | LibreOffice | this library |
|---|---|---|
| elements written | `r, fld, r, fld, r, fld` | `r, r, r, r, r, r` |
| field kinds kept | `slidenum, slidecount, datetime3` | none |
| `slidecount` | live | dead text |
| the date | recomputed | frozen at the stand-in |

## Proposal

A `Field` element in `Shape/RichText/` alongside `Run` and `BreakElement`, carrying the kind of
field, so the Writer emits from the element and the Reader gives one back. `type` is an open
`xsd:string`, not a list: MS reserves `slidenum`, `datetime`, `datetime1`-`datetime13` and
`uaqdatetime1`-`uaqdatetime7` (MS-OI29500 §21.1.2.2.4), and LibreOffice reads more than that --
`slidecount`, `slidename`, `file`, `author` (`oox/source/drawingml/textfield.cxx:105-148`). An
application that does not know a kind shows the text the field last had, which is a graceful
fallback rather than a broken file.

The ODPresentation Writer has the same coupling, and one consequence the PowerPoint2007 one does
not: `getPlaceholderField()` gives a single kind per shape and it is then applied to *every* run in
it, so a slide-number placeholder holding "page " and a number writes both of them as
`text:page-number`. OpenDocument has an element for each kind -- `text:page-number`,
`text:page-count`, `text:date`, `text:time`, `text:author-name`, `text:file-name` -- so the same
map works there, with the fourteen dated formats coming down to a date or a time.

dkulyk:
### Description

A field is a run whose text the application recomputes -- a slide number, a date -- and in both
formats it stands beside the runs around it. Here it was written from the **shape**: the
PowerPoint2007 Writer emitted `a:fld` when the whole shape was a `sldNum` or a `dt` placeholder,
and the ODPresentation Writer took one kind of field per shape and then applied it to every run in
it. So a whole shape could be a slide number, but a field could not sit inside a sentence, and
"page 7 of 12" was inexpressible.

`Shape\RichText\Field` is a run that carries the kind of field, after LibreOffice's
`TextField : TextRun` (`oox/inc/drawingml/textfield.hxx:31`). The writers emit from the element,
among the runs around it, and the readers give a `Field` back.

```php
$paragraph = $slide->createRichTextShape()->getActiveParagraph();
$paragraph->createTextRun('page ');
$paragraph->createField(Field::TYPE_SLIDENUM, '<nr.>');
$paragraph->createTextRun(' of ');
$paragraph->createField(Field::TYPE_SLIDECOUNT, '12');
```

The kind is an open string rather than a list of constants, because `a:fld/@type` is an open
`xsd:string`. `slidenum`, `datetime` and `datetime1`-`datetime13` are the reserved ones;
LibreOffice reads and writes `slidecount`, `slidename`, `file`/`file1`-`file3` and `author` as
well, and an application that does not know a kind shows the text the field stands in for. The
documentation lists them all with what each says.

OpenDocument names a field by what it is rather than by how it is formatted, so the ODPresentation
Writer maps the kinds onto `text:page-number`, `text:page-count`, `text:date`, `text:time`,
`text:author-name` and `text:file-name`, with the fourteen dated formats coming down to a date or a
time; a kind it has no element for is written as the text it stands in for. The Reader is that map
read backwards.

Writing from the shape's placeholder stays as it was, for a placeholder that holds no field of its
own. The Reader half of the PowerPoint2007 side landed in #916; this is the rest of it.

**Checked against LibreOffice** (`939a681ce250`), both directions:

* the PPTX this branch writes converts to `text:page-number` and `text:page-count`;
* the ODP this branch writes converts back to `a:fld type="slidenum"` and `type="slidecount"`.

One note on `testRichTextFieldSlideCount`, which asserts the element but not the schema: libxml
rejects `text:page-count` inside a `text:span` -- along with every other element of the two groups
the schema spells as a single `<element>` holding a choice of names -- while accepting the same
element as a direct child of `text:p`, which has the identical content model. Jing validates the
file, and LibreOffice writes the count inside a span too. `Sample_23_Field.php` puts the construct
in front of the ODF Validator job, which does check it.

Fixes #958

### Checklist:

- [x] My CI is :green_circle:
> 1049 tests green locally with `--fail-on-warning --fail-on-notice --fail-on-deprecation
> --fail-on-phpunit-deprecation`; phpstan, phpmd and php-cs-fixer clean.
> `FieldTest`, `PptSlidesTest::testRichTextField*`, `ContentTest::testRichTextField*`, and a
> round trip through each Reader. Every one of them was checked against the old `src/` and fails
> there.
> `docs/usage/shapes/richtext.md` gains a "Field" section: the API, the table of kinds, and what
> each Writer does with one.
> One enhancement line and one bug-fix line, for the ODPresentation placeholder field being
> written on every run of the shape.


src/PhpPresentation/Shape/RichText/Paragraph.php
```diff
@@ -215,6 +215,21 @@ public function createTextRun(string $pText = ''): Run
...
+     * @param string $type Field type
+     * @param string $pText Text the field last had, for whoever cannot compute it
+     */
+    public function createField(string $type = Field::TYPE_SLIDENUM, string $pText = ''): Field
+    {
+        $objText = new Field($type, $pText);
+        $objText->setFont(clone $this->font);
+        $this->addText($objText);
+
+        return $objText;
```
Progi1984:
```suggestion
        $field = new Field($type, $pText);
        $field->setFont(clone $this->font);
        $this->addText($field);

        return $field;
```

Progi1984 (commented):
A small feedback

(And need rebase)

src/PhpPresentation/Shape/RichText/Paragraph.php
dkulyk: @Progi1984 Applied.

Progi1984 (approved):

PR #961, opened 2026-09-01
PowerPoint2007 Reader : Read back the alpha a colour was written with

### LINKED ISSUE #960: PowerPoint2007 Reader : a colour with an alpha comes back opaque, and a digit short
`loadStyleColor()` reads `a:alpha` back with arithmetic of its own, and gets it wrong twice:

```php
$alpha = strtoupper(dechex((int) (((int) $oElementAlpha->getAttribute('val') / 1000) / 100) * 255));
$oColor->setRGB($oElement->getAttribute('val'), $alpha);
```

`a:alpha` counts in thousandths of a percent, so `val="84000"` is 84%.

1. The `(int)` cast binds to the division, not to the product: `(int) 0.84` is `0`, and `0 * 255` is
   `0`. Every alpha but `FF` is read back as none at all.
2. `dechex(0)` is `"0"` -- one character, not two. `setRGB()` puts it straight in front of the
   colour, so the ARGB comes out seven characters long and every accessor is off by one.

A white plate at 84%, written by this library and read back by it:

```php
$color = new Color('D6FFFFFF');   // written as <a:srgbClr val="FFFFFF"><a:alpha val="84000"/>
// after the round trip
$color->getARGB();   // '0FFFFFF'  -- seven characters
$color->getRGB();    // 'FFFFF'    -- the first digit went into the alpha
$color->getAlpha();  // 100        -- the transparency is gone
```

Only `FF` survives: `(int) ((100000 / 1000) / 100)` is `1`, and `dechex(255)` is two characters.

This is not confined to one kind of shape. `loadStyleColor()` is what every colour in the
PowerPoint2007 Reader goes through -- solid, gradient and pattern fills, outlines, the fill of a
chart serie and of its data points.

## Suggested fix

`Color::setAlpha()` already does this arithmetic, and does it correctly -- it rounds rather than
truncates, and pads to two characters. The reader can hand it the percent and let it do the work:

```php
$oColor->setAlpha((int) round((int) $oElementAlpha->getAttribute('val') / 1000));
```

One thing to settle with it: `setAlpha()` writes the hex pair in lower case, in front of a colour
that is upper case everywhere else in the library, including the `Color::COLOR_*` constants. Worth
upper-casing there rather than working around it at the call site.

dkulyk:
### Description

`loadStyleColor()` did its own arithmetic on `a:alpha` and got it wrong twice. The `(int)` cast bound
to the division rather than to the product, so `(int) 0.84` was `0` and every alpha but `FF` was read
back as none at all; and `dechex(0)` is one character, so the ARGB came out seven characters long and
the colour lost its first digit to the alpha.

```php
$color = new Color('D6FFFFFF');   // <a:srgbClr val="FFFFFF"><a:alpha val="84000"/>
// after a round trip through the reader
$color->getARGB();   // '0FFFFFF'  -- seven characters
$color->getRGB();    // 'FFFFF'    -- the first digit went into the alpha
$color->getAlpha();  // 100        -- the transparency is gone
```

`Color::setAlpha()` already does that arithmetic, and does it correctly: it rounds rather than
truncates, and pads to two characters. The reader hands it the percent and lets it do the work. It
wrote the hex pair in lower case in front of a colour written in upper case everywhere else,
including the `COLOR_*` constants, so it upper-cases it now.

Every colour in this reader goes through that one method -- solid, gradient and pattern fills,
outlines, the fill of a chart serie and of its data points -- so this is not confined to one kind of
shape.

Fixes #960

### Checklist:

- [x] My CI is :green_circle:
> Green locally: 1046 tests, phpstan, phpmd and php-cs-fixer all clean.
> `testColorAlphaIsReadBack` round-trips three colours through the writer and the reader: an opaque
> one, the 84% a data label plate is written with, and `0A`, whose alpha is a single hex digit and
> has to keep its leading zero.
> Nothing to describe: this is a fix to reading back what the library already wrote, and no API
> changed.
> One line under Bug fixes in `docs/changes/1.3.0.md`.

### DIFF
--- a/tests/PhpPresentation/Tests/Reader/PowerPoint2007Test.php
+++ b/tests/PhpPresentation/Tests/Reader/PowerPoint2007Test.php
@@ -1254,6 +1254,51 @@ class PowerPoint2007Test extends TestCase
         self::assertEquals(Fill::FILL_NONE, $seriesRead->getDataPointOutline(1)->getFill()->getFillType());
     }
 
+    /**
+     * @dataProvider dataProviderColorAlpha
+     */
+    #[DataProvider('dataProviderColorAlpha')]
+    public function testColorAlphaIsReadBack(string $argb, int $expectedAlpha): void
+    {
+        $oPhpPresentation = new PhpPresentation();
+        $oSeries = new Series('Downloads', ['Jan' => '1', 'Feb' => '5', 'Mar' => '2']);
+        $oSeries->getFill()->setFillType(Fill::FILL_SOLID)->setStartColor(new Color($argb));
+        $oBar = new Bar();
+        $oBar->addSeries($oSeries);
+        $oPhpPresentation->getActiveSlide()->createChartShape()->getPlotArea()->setType($oBar);
+
+        $file = tempnam(sys_get_temp_dir(), 'PhpPresentation');
+        (new PowerPoint2007Writer($oPhpPresentation))->save($file);
+        $oPhpPresentationRead = (new PowerPoint2007())->load($file);
+        unlink($file);
+
+        $arrayShape = array_values((array) $oPhpPresentationRead->getActiveSlide()->getShapeCollection());
+        self::assertInstanceOf(Chart::class, $arrayShape[0]);
+        $seriesRead = $arrayShape[0]->getPlotArea()->getType()->getSeries();
+        $seriesRead = reset($seriesRead);
+
+        // the colour comes back whole: an alpha that is not `FF` used to be read as `0`, and as a
+        // single character it took the first digit of the colour with it
+        self::assertEquals($argb, $seriesRead->getFill()->getStartColor()->getARGB());
+        self::assertEquals(substr($argb, 2), $seriesRead->getFill()->getStartColor()->getRGB());
+        self::assertEquals($expectedAlpha, $seriesRead->getFill()->getStartColor()->getAlpha());
+    }
+
+    /**
+     * @return array<array{string, int}>
+     */
+    public static function dataProviderColorAlpha(): array
+    {
+        return [
+            // ARGB written, and the percent it is read back as
+            ['FFFF7700', 100],
+            // 84%: the plate a data label sits on
+            ['D6FFFFFF', 84],
+            // an alpha of one hex digit, which has to keep its leading zero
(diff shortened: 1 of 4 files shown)


tests/PhpPresentation/Tests/Reader/PowerPoint2007Test.php
```diff
@@ -1254,6 +1254,49 @@ public function testSeriesDataPoints(): void
...
+        // the colour comes back whole: an alpha that is not `FF` used to be read as `0`, and as a
+        // single character it took the first digit of the colour with it
+        self::assertEquals($argb, $seriesRead->getFill()->getStartColor()->getARGB());
+        self::assertEquals(substr($argb, 2), $seriesRead->getFill()->getStartColor()->getRGB());
+    }
+
+    /**
+     * @return array<array<string>>
+     */
+    public static function dataProviderColorAlpha(): array
```
Progi1984, line 1290: Could you add to this dataProvider the expected alpha ?

dkulyk: @Progi1984 Added: the provider now carries the percent each colour is read back as -- 100, 84 and 4 -- and the test asserts `getAlpha()` against it.

dkulyk:
@Progi1984 The red ODF Validator is not this branch: Maven Central answered `429 Too Many Requests` when the job fetched the validator jar, before any document was validated.

```
--2026-09-02 07:06:58--  https://repo1.maven.org/maven2/org/odftoolkit/odfvalidator/0.12.0/odfvalidator-0.12.0-jar-with-dependencies.jar
HTTP request sent, awaiting response... 429 Too Many Requests
ERROR 429: Too Many Requests.
```

The same job was green on the previous push of this branch, and the only change since is the expected alpha you asked for in the data provider. Everything else here is green. I have no rights to rerun it and the workflow has no `workflow_dispatch`, so it needs a rerun from your side if you want it green before merging -- I did not want to force-push just to retrigger CI.

Separately, that step is a bare `wget` with no retry, so it will fail this way for anyone whenever Maven Central rate-limits. Two ways to remove it: `--tries`/`--waitretry` on the wget, or caching the jar with `actions/cache`.

Progi1984 (approved):

PR #963, opened 2026-09-01
Shape\Chart\Series : Give the data labels a fill of their own

### LINKED ISSUE #962: Shape\Chart\Series : a data label cannot be given a fill to sit on
A data label is drawn straight over the chart, so it is read against whatever colour the slice under
it happens to be. PowerPoint answers this with a plate behind the text -- a `c:spPr` inside
`c:dLbls`, usually white and semi-transparent:

```xml
<c:dLbls>
  <c:spPr>
    <a:solidFill><a:schemeClr val="bg1"><a:alpha val="84000"/></a:schemeClr></a:solidFill>
    <a:ln><a:noFill/></a:ln>
  </c:spPr>
  …
```

The library cannot say this. `Series` has `fill`, `dataPointFills`, `font` and `outline`, and all of
them are about the slices rather than the labels, and none of the nine `writeType*` methods in
`PptCharts` writes a `c:spPr` inside `c:dLbls` at all. A presentation read in with such a plate loses
it on save.

## Suggested fix

- `Series::getLabelFill()`, holding a `Fill` born of type `Fill::FILL_UNSET` the way a fill nobody
  named is held everywhere else, so that a serie which does not ask writes no `c:spPr` and nothing
  changes for a file that never had one. `FILL_NONE` stays a different answer -- the labels sit on
  nothing, and it writes `a:noFill`.
- One helper in `PptCharts`, called from all nine `writeType*` methods after `c:numFmt` and before
  `c:txPr`, which is where `CT_DLbls` puts `c:spPr`.
- The reader reads it back beside the `c:spPr` of the serie it already reads.

## There is no ODP side to this

Worth recording, because it looks like an omission. ODF 1.2 has no way to say what a data label sits
on. LibreOffice invented one in its own namespace, on the chart style of the serie:

```xml
<style:chart-properties loext:label-fill="solid" loext:label-fill-color="#ffffff" …/>
```

`jing` against the OpenDocument 1.2 schema rejects it -- `attribute "loext:label-fill-color" not
allowed here` -- so writing it would fail this project's own `assertIsSchemaOpenDocumentValid('1.2')`
and the ODF Validator job. LibreOffice also drops the transparency on the way out: a plate at 84%
comes back plain white. The ODPresentation Reader reads no charts at all, so there is nothing on that
side to read it back with either.

dkulyk:
### Description

A data label is drawn straight over the chart, so it is read against whatever colour the slice under
it happens to be. PowerPoint answers this with a plate behind the text -- a `c:spPr` inside
`c:dLbls`. `Series` had no field for it (`fill`, `dataPointFills`, `font` and `outline` are all about
the slices) and none of the nine `writeType*` methods wrote a `c:spPr` at all, so a presentation read
in with such a plate lost it on save.

```php
$series->getLabelFill()
    ->setFillType(Fill::FILL_SOLID)
    ->setStartColor(new Color('D6FFFFFF'));
```

```xml
<c:dLbls>
  <c:spPr><a:solidFill><a:srgbClr val="FFFFFF"><a:alpha val="84000"/></a:srgbClr></a:solidFill></c:spPr>
  <c:txPr>…
```

The fill is born of type `Fill::FILL_UNSET`, the way a fill nobody named is held everywhere else, so
a serie that does not ask writes no `c:spPr` and nothing changes for a file that never had one.
`FILL_NONE` stays a different answer: the labels sit on nothing, and it writes `a:noFill`. One helper
writes the element at all nine call sites, after `c:numFmt` and before `c:txPr`, which is where
`CT_DLbls` puts it. The reader reads it back beside the `c:spPr` of the serie it already reads.

Checked against LibreOffice in both directions: it reads our `c:dLbls/c:spPr` and re-emits it per
label as `c:dLbl/c:spPr`, keeping the colour and the transparency.

**There is no ODP side to this**, which is worth saying because it looks like an omission. ODF 1.2
has no way to express what a data label sits on; LibreOffice invented `loext:label-fill` and friends
in its own namespace, `jing` rejects those against the OpenDocument 1.2 schema, and LibreOffice
itself drops the transparency when it writes them. The ODPresentation Reader reads no charts at all,
so there is nothing on that side to read it back with either.

This does not depend on #961 and the two can land in either order. The writer test asserts the
`a:alpha` that a semi-transparent plate is written with, which has always been written correctly;
the reader test uses an opaque plate on purpose, because a colour with an alpha does not survive
this reader at all -- that is #960, and #961 fixes it.

Fixes #962

### Checklist:

- [x] My CI is :green_circle:
> Green locally: 1047 tests, phpstan, phpmd and php-cs-fixer all clean.
> `testSeriesDataLabelFill` runs over all nine chart types through the existing data provider and
> asserts the schema each time, `testSeriesLabelFillIsReadBack` round-trips one through the reader,
> and `SeriesTest::testLabelFill` covers the `FILL_UNSET` default.
> A "Label Fill" section under Series in `docs/usage/shapes/chart.md`.
> One line under Enhancements in `docs/changes/1.3.0.md`.

### DIFF
--- a/docs/changes/1.3.0.md
+++ b/docs/changes/1.3.0.md
@@ -13,6 +13,7 @@
 - Added `Table::setFirstRow()` and `Table::setBandRow()`, so that a table can say whether its first row is a header row, by [@dkulyk](http://github.com/dkulyk) in [#904](https://github.com/PHPOffice/PHPPresentation/pull/904)
 - PowerPoint2007 Reader : Read a group of shapes, and map the shapes it holds out of the coordinate space `a:chOff`/`a:chExt` declares, by [@Wilkolicious](http://github.com/Wilkolicious) in [#870](https://github.com/PHPOffice/PHPPresentation/pull/870) and [@dkulyk](http://github.com/dkulyk) in [#947](https://github.com/PHPOffice/PHPPresentation/pull/947)
 - Added `Shape\RichText\Field`, so that a slide number, a slide count or a date can stand inside a sentence rather than take a whole shape, written and read back by the PowerPoint2007 Writer and Reader and by the ODPresentation Writer and Reader, by [@dkulyk](http://github.com/dkulyk) fixing [#958](https://github.com/PHPOffice/PHPPresentation/issues/958) in [#959](https://github.com/PHPOffice/PHPPresentation/pull/959)
+- PowerPoint2007 Writer & Reader : Added a fill behind the data labels of a chart serie, so that a label stays readable over the chart it sits on, by [@dkulyk](http://github.com/dkulyk) in [#963](https://github.com/PHPOffice/PHPPresentation/pull/963)
 
 ## Bug fixes
 - PowerPoint2007 Reader : Fixed a colour with an alpha coming back opaque and one character short, taking the first digit of the colour with it, by [@dkulyk](http://github.com/dkulyk) in [#961](https://github.com/PHPOffice/PHPPresentation/pull/961)
--- a/docs/usage/shapes/chart.md
+++ b/docs/usage/shapes/chart.md
@@ -344,6 +344,20 @@ $series = new Series('Downloads', $seriesData);
 $series->setLabelPosition(Series::LABEL_INSIDEEND);
 ```
 
+#### Label Fill
+You can draw a plate behind the data labels, so that they stay readable over the chart.
+A series that names no label fill leaves the labels as the application draws them.
+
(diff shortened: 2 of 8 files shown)


Progi1984 (commented):
Need a rebase ;)

dkulyk:
@Progi1984 Rebased onto `master`, green at 25/25.

Progi1984 (approved):

PR #964, opened 2026-09-02
ODPresentation Writer : Share the table row and the table cell styles too

dkulyk:
### Description

[#935](https://github.com/PHPOffice/PHPPresentation/pull/935) put the paragraph and the text families in a style pool, [#936](https://github.com/PHPOffice/PHPPresentation/pull/936) the graphic and the drawing-page ones. The two table families were the remainder, and they still named a style after the *position* of the thing wearing it — `gr` plus the shape counter, the row index and the cell index. A positional name cannot be shared, so a table wrote one definition per row and one per cell however alike they were.

Five slides holding a five-by-four table, every row and every cell alike:

| | `master` | this PR |
| --- | --- | --- |
| `style:family="table-row"` definitions | 25 | **1** |
| `style:family="table-cell"` definitions | 100 | **1** |
| `content.xml` | 46 893 bytes | **27 834 bytes** |

The names are `ro1`, `ce1` — the prefixes LibreOffice uses for the two families.

**The pool shares what is alike and nothing else.** The same five slides, but with five row heights and four cell fills between them:

| | `master` | this PR |
| --- | --- | --- |
| `table-row` definitions | 25 | **5** |
| `table-cell` definitions | 100 | **4** |
| `content.xml` | 54 693 bytes | **29 161 bytes** |

**Nothing about the file changes but the names and the count.** Resolving every `table:style-name` in document order to the full body it points at, on both documents:

| | references | identical | different | dangling |
| --- | --- | --- | --- | --- |
| uniform | 125 | **125** | 0 | 0 |
| varied | 125 | **125** | 0 | 0 |

### Why the callable form of `shareAutomaticStyle()` was needed

A cell's style body is read out of **two** objects: the cell, and the row it sits in, because a cell that named no fill of its own is painted with the row's. A closure captures both. The string form `master` carried before #936 would have had to carry the resolved fill under a key of its own and then throw it away. This is the reason the table half waited for #936 rather than being stacked on it.

`writeTableStyle()` is `addTableStyle()` here, since it writes nothing itself any more — the same rename `writeTxtStyle()`, `writeDrawingStyle()` and `writeLineStyle()` took in #936. The `Styles` writer keeps its own unrelated `writeTableStyle()`.

A side effect worth naming: `$this->shapeId` is no longer read for a table name. Its only remaining reader is `writeShapeChart()`, which writes `./Object N` from it — so the two passes still have to count in step, but only for the charts.

### The test that was missing

`table:table-row/@table:style-name` was the one generated style reference nothing in the suite held the Writer to. Measured on `master` by replacing each reference with a literal:

| reference broken | `ODPresentation/` tests failing |
| --- | --- |
| `table:table-cell/@table:style-name` | 16 |
| `table:table-row/@table:style-name` | **0** |

That is the gap [#934](https://github.com/PHPOffice/PHPPresentation/pull/934) closed everywhere else. It fails **2** now.

Verified in both directions: breaking the row reference fails those two, and renaming the definition alone — `ro` to `RENAMEDro`, the reference following it — leaves the suite green, which is the point. The tests resolve the name from the reference the document carries rather than spelling one out.

Three tests, no existing one changed:

- a `table-row` style reached through its reference, asserting the family and the row height;
- four alike rows of three alike cells are one definition each, and every one of the sixteen references names it;
- two heights and four colours are two definitions and four.

### Checklist:

- [x] My CI is :green_circle:
  > 1118 tests, 9223 assertions, PHPStan level 4, PHPMD and PHP-CS-Fixer all clean locally.
  > Three new tests, described above. No existing assertion had to change: with #934 in place every assertion about a generated style already resolves the name from the reference the document carries, so renaming `gr{shapeId}r{n}` to `ro1` and `gr{shapeId}r{n}c{m}` to `ce1` is invisible to the suite — 197 of 197 `ODPresentation/` tests passed on the first run after the change.
  > Nothing to update: no public API changes. What changes is the name a generated style is given and how many are written, neither of which the documentation describes.
  > One line under Bug fixes, added once this pull request had a number.


Progi1984 (commented):
Need a rebase

dkulyk:

dkulyk:

dkulyk:

Progi1984 (commented):
@dkulyk Need a rebase

dkulyk:

Progi1984 (approved):

PR #966, opened 2026-09-02
Presentation : Add moveSlide(), and say how a slide moves

### LINKED ISSUE #477: Is there any way by which we can move slide to a particular index
I am trying to figure if there is a way to insert a slide at a particular index instead of the default behavior of appending it to the end? (for e.g when using createSlide() can I specify the index where it has to be inserted. 
Or is there a way to move a slide to a particular index in the presentation?

dkulyk:
### Description

`addSlide()` has taken a position since #810, and `docs/usage/slides/introduction.md` documents
that. Moving a slide that is **already** in the deck is the other half of the same question, and
the reporter of #477 asked for both. Until now the second one had no method and no page: a caller
had to find `getIndex()`, `removeSlideByIndex()` and `addSlide()` and get the order right.

`PhpPresentation::moveSlide(Slide $slide, int $index)` does it in one call.

```php
$presentation->moveSlide($slide, 2);
```

The index counts in the collection *without* the slide being moved, which is the part worth writing
down: in a presentation of `A B C D`, moving `A` to 2 gives `B C A D`, not `A` landing where it
stood before the removal. It throws `OutOfBoundsException` for an index past the last slide and
`InvalidParameterException` for a slide the presentation does not hold.

One signature rather than two: a union `Slide|int $slide` would let a caller pass either, but a
native union type is PHP 8.0 and the floor is `^7.4`, and `getSlide()` already turns an index into
a slide. `moveSlideByIndex(int $from, int $to)` was the alternative and reads ambiguously -- both
parameters are indices and the name says which is which only by position.

This also fixes the paragraph above the new section, which called the parameter of `addSlide()`
`$position`; the signature calls it `$index`.

Fixes #477

### Checklist:

- [x] My CI is :green_circle:
> 1121 tests / 9152 assertions green locally, with `--fail-on-warning --fail-on-notice
> --fail-on-deprecation --fail-on-phpunit-deprecation`. phpstan, phpmd and php-cs-fixer clean.
> Six tests in `PhpPresentationTest`: moving earlier, moving later (the index-shift case),
> moving a slide to where it already is, the fluent return, a slide the presentation does not hold,
> and an index past the last slide.
> `docs/usage/slides/introduction.md` -- a `Move a slide` section, next to the one about adding at
> a position.
> `docs/changes/1.3.0.md`.


Progi1984 (commented):
Need a rebase

dkulyk:

Progi1984 (approved):

Issue #943, opened 2026-08-26
`Border` and `Outline` are two model classes for one line, and neither can describe it


dkulyk:
A design question rather than a bug report, and it wants a maintainer's decision before anything is
written. Three defects and three user reports below all follow from one split.

Re-measured against `master` on 2 Sep 2026; every figure and line number is current.

## Two classes for one `a:ln`

`Style\Border` and `Style\Outline` both describe **the stroke of a shape**, and both are written as
`a:ln` in PowerPoint2007 (`AbstractDecoratorWriter.php:63` and `:246`) and as `draw:stroke` /
`svg:stroke-*` in ODPresentation. Which class an object gets is decided by history — `Outline` was
added in c25b553 (2016, #169) for a chart series, when `Border` already carried a width, a colour, a
dash and a compound style — not by the model:

| carries `Border` | carries `Outline` |
| --- | --- |
| `AbstractShape` — so every shape | `Shape\AutoShape` |
| `Shape\Chart\Legend` | `Shape\Chart\Axis` |
| `Shape\Chart\Marker` | `Shape\Chart\Gridlines` |
| a table `Cell` (four of them, via `Borders`) | `Shape\Chart\Series` |

## Neither class covers `CT_LineProperties`

- `Border` has `lineWidth`, `lineStyle`, `dashStyle`, `Color`. It **cannot** describe a gradient or
  pattern line: its colour is a `Color`, not a `Fill`.
- `Outline` has `width` and a `Fill`. It **cannot** describe a dash or a compound line, having
  neither field.

Between them they cover the type; separately, neither does. LibreOffice keeps the whole of it in one
struct, `oox/inc/drawingml/lineproperties.hxx`:

```cpp
struct LineProperties {
    FillProperties      maLineFill;      // what this library calls Outline
    DashStopVector      maCustomDash;
    std::optional<sal_Int32> moLineWidth, moPresetDash, moLineCompound, moLineCap, moLineJoint;
};
```

`LineProperties::getLineStyle()` reads both halves to answer one question: whether a line exists is a
property of the fill — our `Outline`; whether it is solid or dashed is a property of the dash — our
`Border`. Neither PHP class can see the other.

## What follows from the split

**1. A compound `Line` shape is written to ODP as no line at all.** `addLineStyle()`,
`Content.php:1416`, maps everything that is not `LINE_SINGLE` to `draw:stroke="none"` — so
[...]

PR #974, opened 2026-09-02
ODPresentation Writer and Reader : Put a turned shape where it was asked to be, and read it back

### LINKED ISSUE #973: ODPresentation : A rotated shape is written in the wrong place, and its position is never read back
### Description

A `RichText` shape that carries a rotation is written into `content.xml` as a `draw:transform` rather than as `svg:x`/`svg:y`. Two things go wrong with it.

**1. The shape is written in the wrong place.** `translate` moves the frame *after* `rotate` has turned it about the origin, while the rotation a shape carries is about its own centre. The point to write is therefore where the top left corner lands once the shape has been turned. `Writer/ODPresentation/Content.php` writes the raw offsets instead — and writes them transposed, X into Y and Y into X.

**2. Nothing reads it back.** `Reader/ODPresentation.php` looks only at `svg:x`/`svg:y`, so a turned shape comes back at (0, 0) with no rotation, whichever loader picks it up.

The first half is visible without any reader involved: an ODP written today puts a rotated shape somewhere else than the PPTX written from the same deck.

### Steps to reproduce

```php
$shape->setOffsetX(400)->setOffsetY(100)->setWidth(200)->setHeight(100)->setRotation(30);
$shape->createTextRun('Turned');
```

### Expected behavior

The shape renders where the same deck renders through the PowerPoint2007 Writer, and reads back with the offset and the rotation it was given.

### Actual behavior

Written:

```xml
draw:transform="rotate (-0.5235987755983) translate (2.646cm 10.583cm)"
```

2.646cm is the Y offset and 10.583cm is the X one. LibreOffice, saving the same shape, writes `translate (11.599cm 1.5cm)`.

Rendered through LibreOffice 26.2 and measured on the shape's fill, the ODP puts its centre at (214, 653) where the PPTX from the same deck puts it at (661, 198). The bounding boxes match at 293×245, so the rotation itself is applied — only the placement is wrong.

Read back with `IOFactory::createReader('ODPresentation')`, the shape returns `getOffsetX() === 0`, `getOffsetY() === 0`, `getRotation() === 0`.

### Notes

Fixing the Writer moves a rotated shape in newly written files. Files written by earlier versions keep the position they were given.

dkulyk:
### Description

Fixes #973

A turned shape is written as `draw:transform="rotate (...) translate (...)"` instead of `svg:x`/`svg:y`. `translate` moves the frame after `rotate` has turned it about the origin, while a shape's own rotation is about its centre, so the point to write is where the top left corner lands once the shape has been turned. The Writer wrote the raw offsets, transposed into the bargain, and the Reader read neither.

**Writer.** `translate` is now the turned corner:

```
tx = x + w/2 - (w/2)·cos θ + (h/2)·sin θ
ty = y + h/2 - (w/2)·sin θ - (h/2)·cos θ
```

For the shape in the issue that gives `translate (11.599cm 1.500cm)`, which is byte for byte what LibreOffice writes for it. Rendered and measured on the fill, the ODP and the PPTX from the same deck now land on the same centre (661, 198) with the same 293×245 box; before, the ODP was at (214, 653).

**Reader.** The three loaders that read a frame's position each carried the same pair of `svg:x`/`svg:y` lines. They now call one `loadShapeOffset()`, which turns the corner back out of a `draw:transform` when the frame has one and falls back to `svg:x`/`svg:y` when it does not. Rotation is read for the first time. Only the `rotate`/`translate` form is parsed, which is what this Writer and LibreOffice write; any other transform list leaves the shape where the `svg` attributes put it, as before.

A rotated shape in a newly written file therefore moves to where it was asked to be. Files written by earlier versions are not touched.

Found by walking every property of a shape through each Writer/Reader pair and diffing what came back; this closes three of the differences the ODP pair showed.

### Checklist:

> A Writer test pinning `translate` against the value LibreOffice writes, and a Reader round trip over 0°, 30°, 90°, 145° and 270°. Five of the six fail on `develop`.
> `docs/changes/1.3.0.md`.
> Same file.


Progi1984 (commented):
Need a rebase

dkulyk:
@Progi1984 Rebased onto `master`. More will follow: 16 more gaps are written up and measured on my side. 14 of them can go in one pull request each, and two wait on #943.

Progi1984 (approved):

PR #980, opened 2026-09-02
ODPresentation Reader : Read a paragraph's line spacing out of the line height

### LINKED ISSUE #979: ODPresentation Reader : A paragraph's line spacing is read out of the margin below it
### Description

`ODPresentation::loadStyle()` reads `fo:line-height` -- it asks whether the attribute is there -- and then takes both the unit and the value out of `fo:margin-bottom` instead. So a paragraph's line spacing comes back as the margin below it, converted from centimetres and cut to an integer, and the mode always reads as points because a margin is written in centimetres and never in per cent.

Beside it, `fo:margin-top` and `fo:margin-bottom` are written with three decimals of a centimetre. A point is 127/3600 cm, so three decimals lose about a seventieth of a point and a spacing does not come back whole.

### Steps to reproduce

```php
$shape->createTextRun('Sample');
$shape->getActiveParagraph()
    ->setLineSpacingMode(Paragraph::LINE_SPACING_MODE_PERCENT)
    ->setLineSpacing(150)
    ->setSpacingBefore(11)
    ->setSpacingAfter(13);
```

### Expected behavior

Reading the file back gives 150 per cent, 11 points before and 13 after.

### Actual behavior

`getLineSpacing()` is `0` and `getLineSpacingMode()` is `point`: the margin below the paragraph is `0.459cm`, whose unit is not `%`, and whose value cut to an integer is 0. `getSpacingBefore()` is 10.998425196850395 and `getSpacingAfter()` is 13.011023622047245.

### Notes

`Alignment::getLevel()` is lost as well and is not part of this. A level lives in the nesting of `text:list`, and a list is written only for a paragraph that carries a bullet, so a level on a plain paragraph has nowhere to go. Where it should go is a decision rather than a fix.

dkulyk:
### Description

Fixes #979

`loadStyle()` asks whether `fo:line-height` is there and then takes both the unit and the value out of `fo:margin-bottom`. A line spacing therefore came back as the margin below the paragraph, cut to an integer, and the mode always came back as points, because a margin is written in centimetres and never in per cent.

`fo:margin-top` and `fo:margin-bottom` are written to six decimals rather than three: a point is 127/3600 cm, so three decimals lose about a seventieth of a point, and 11 points came back as 10.998425196850395. The two Writer tests that pinned the old three decimals now pin the six.

`Alignment::getLevel()` is lost as well and is deliberately left alone: a level lives in the nesting of `text:list`, and a list is written only for a paragraph that carries a bullet, so a level on a plain paragraph has nowhere to go. Where it should go is a decision rather than a fix.

Found by walking every property of a shape through each Writer/Reader pair and diffing what came back; this closes four of the differences the ODP pair showed.

### Checklist:

> A round trip over the line spacing in both modes and one over the two paragraph spacings. All three fail on `develop`.
> `docs/changes/1.3.0.md`.
> Same file.

### DIFF
--- a/docs/changes/1.3.0.md
+++ b/docs/changes/1.3.0.md
@@ -76,6 +76,7 @@
 - ODPresentation Writer : Fixed the paragraph, the text and the graphic style of a slide note being referenced but never written, and shared one automatic style between everything that writes the same one, by [@dkulyk](http://github.com/dkulyk) in [#935](https://github.com/PHPOffice/PHPPresentation/pull/935)
 - ODPresentation Writer : Shared the graphic and the drawing-page styles between everything that writes the same one, the way the paragraph and the text ones are, by [@dkulyk](http://github.com/dkulyk) in [#936](https://github.com/PHPOffice/PHPPresentation/pull/936)
 - PowerPoint2007 Reader : Fixed a shadow's alignment being read off the effect list rather than off the shadow in it, and its colour and alpha being looked for under a preset colour the Writer never writes, and read a shape's wrap back, by [@dkulyk](http://github.com/dkulyk) in [#982](https://github.com/PHPOffice/PHPPresentation/pull/982)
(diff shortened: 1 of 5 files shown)


Progi1984 (commented):
Need a rebase :)

Progi1984 (approved):

dkulyk:

PR #970, opened 2026-09-02
PowerPoint2007 Reader : Read back the border a shape was written with

dkulyk:
### Description

The Writer writes a shape's `Border` as `a:ln` in `p:spPr`, from `writeShapeText()` (`AbstractSlide.php:270`), `writeShapeLine()` (`:872`) and `writeShapePic()` (`:1568`). The Reader opened the same `p:spPr`, took the fill out of it and stopped -- so a shape loaded from a file always wore the border it was born with (`LINE_NONE`, `AbstractShape.php:145`), whatever the file said.

A `RichText` given a black single 1px border, written, loaded and written again:

```xml
<!-- written, and after this PR also on the way back -->
<a:ln w="9525" cap="flat" cmpd="sng" algn="ctr">
  <a:solidFill><a:srgbClr val="000000"><a:alpha val="100000"/></a:srgbClr></a:solidFill>
  <a:prstDash val="solid"/><a:round/>
  ...
</a:ln>

<!-- loaded, then written again -- before this PR -->
<!-- nothing -->
```

No new parsing code: `loadStyleBorder()` (`Reader/PowerPoint2007.php:1923`) already reads exactly that element -- `w` through `emuToPixels()`, `cmpd` into the line style, `a:noFill` into `LINE_NONE`, `a:solidFill/a:srgbClr` into the colour, `a:prstDash` into the dash. It was only ever called for the six borders of a table cell (`:1330`-`:1350`). The two shape loaders now call it as well, on the `p:spPr` they already have open.

The Reader creates no `Line` shapes, so `writeShapeLine()` has no counterpart to fix. The ODPresentation Reader has the same gap -- `Border` reaches its model only through `$oCell->setBorders()` -- and is not touched here.

Fixes #969

### Checklist:

- [x] My CI is :green_circle:
> 1117 tests / 9155 assertions green locally with `--fail-on-warning --fail-on-notice --fail-on-deprecation --fail-on-phpunit-deprecation`. phpstan, phpmd and php-cs-fixer clean. No existing test changed behaviour.
- [x] I have covered by unit tests my new code (check build/coverage for coverage report)
> Two round-trip tests in `Reader/PowerPoint2007Test`, one per `p:spPr` call site: a `RichText` whose width, compound, dash and colour all come back, and a `Drawing\Gd` that goes through the other loader. Both fail on `master` -- the width comes back as the default `1.0` instead of the `4` and the `2` that were written.
[...]

PR #968, opened 2026-09-02
PowerPoint2007 Reader : Read a text inset in the unit the model holds it in

dkulyk:
### Description

The text insets of a shape are EMU in the file and pixels in the model. The Writer converts on the way out (`AbstractSlide.php:301-304`, `pixelsToEmu()`); the Reader stored the attribute as it stood (`Reader/PowerPoint2007.php:1174-1185`), so each round trip multiplied every inset by the 9525 EMU of a pixel.

The defaults show it without a deck being needed. `RichText` is born with `insetBottom = 4.8` and `insetLeft = 9.6` -- pixels -- which the Writer writes as `bIns="45720"`, `lIns="91440"`:

| | `bIns` | `lIns` |
| --- | --- | --- |
| written | 45720 | 91440 |
| loaded, then written again -- before | **435483000** | **870966000** |
| loaded, then written again -- after | 45720 | 91440 |

435483000 / 45720 = 9525 exactly, and a third pass multiplied it again.

The fix is `CommonDrawing::emuToPixels()` on the four reads, the inverse of what the Writer already does.

Fixes #967

### Checklist:

- [x] My CI is :green_circle:
> 1117 tests / 9155 assertions green locally with `--fail-on-warning --fail-on-notice --fail-on-deprecation --fail-on-phpunit-deprecation`. phpstan, phpmd and php-cs-fixer clean.
- [x] I have covered by unit tests my new code (check build/coverage for coverage report)
> Two round-trip tests in `Reader/PowerPoint2007Test`: four insets set by hand come back as the numbers that went in, and the untouched defaults come back as 9.6 / 4.8 rather than as 91440 / 45720.
- [x] I have updated the [documentation](https://github.com/PHPOffice/PHPPresentation/tree/develop/docs) to describe the changes
> No documented behaviour changes -- the insets were always meant to be pixels, and `docs/usage/shapes/richtext.md` does not put a unit on them.
- [x] I have updated the [changelog](https://github.com/PHPOffice/PHPPresentation/blob/develop/docs/changes/1.1.0.md)
> `docs/changes/1.3.0.md`.

PR #982, opened 2026-09-02
PowerPoint2007 Reader : Give a shadow its colour, its alpha and its alignment, and a shape its wrap

### LINKED ISSUE #981: PowerPoint2007 Reader : A shadow's colour, alpha and alignment, and a shape's wrap, are not read back
### Description

Three properties of a shape do not come back from a PPTX.

- `loadShadow()` reads `algn` off `$node`, the `a:effectLst`, instead of off the `a:outerShdw` inside it, so a shadow's alignment always comes back empty.
- The same method looks for the colour under `a:prstClr` only. The Writer writes it as `a:srgbClr`, with the shadow's alpha as the `a:alpha` inside it, so neither the colour nor the alpha is ever read.
- `wrap` is written on `a:bodyPr` beside the insets and the two other attributes the Reader takes from there, and nothing reads it back.

### Steps to reproduce

```php
$shape->createTextRun('Sample');
$shape->setWrap(RichText::WRAP_NONE);
$shape->getShadow()
    ->setVisible(true)
    ->setAlignment(Shadow::SHADOW_BOTTOM_RIGHT)
    ->setColor(new Color('FF00FF00'))
    ->setAlpha(40);
```

### Expected behavior

Reading the file back gives a shadow to the bottom right, green, at 40 per cent, on a shape whose text does not wrap.

### Actual behavior

`getAlignment()` is `''`, `getColor()->getARGB()` is `FF000000`, `getAlpha()` is `50` and `getWrap()` is `square` -- the defaults of a fresh shape and a fresh shadow.

### Notes

`hasColumnsRTL()` comes back `false` where it was `null`, and that is not a bug to fix here: `null` means nobody said, the Writer always writes `rtlCol`, so the file cannot say nothing.

dkulyk:
### Description

Fixes #981

`loadShadow()` read `algn` off `$node`, the `a:effectLst`, instead of off the `a:outerShdw` inside it, so a shadow's alignment always came back empty. The same method looked for the colour under `a:prstClr` only, while the Writer writes it as `a:srgbClr` with the shadow's alpha as the `a:alpha` inside it, so neither the colour nor the alpha was ever read. Both spellings are read now, and the alpha goes to the shadow rather than to the colour, which is where the Writer takes it from.

`wrap` is written on `a:bodyPr` beside the insets and the two other attributes the Reader takes from there, and nothing read it back. `none` and `square` are the two values a shape's wrap holds, spelled the same way in the file.

`hasColumnsRTL()` is deliberately left as it is: `null` means nobody said, and the Writer always writes `rtlCol`, so a file cannot say nothing.

Found by walking every property of a shape through each Writer/Reader pair and diffing what came back; this closes four of the differences the PPTX pair showed. With #968 and #970 it leaves that `null` as the only one.

### Checklist:

> A round trip over the shadow's alignment, colour and alpha, and one over the wrap. Both fail on `develop`.
> `docs/changes/1.3.0.md`.
> Same file.


tests/PhpPresentation/Tests/Reader/PowerPoint2007Test.php
```diff
@@ -1967,4 +1968,50 @@ public function testFieldIsReadBack(): void
...
+
+    public function testShapeShadowSurvivesTheRoundTrip(): void
+    {
+        $oPhpPresentation = new PhpPresentation();
+        $oShape = $oPhpPresentation->getActiveSlide()->createRichTextShape();
+        $oShape->createTextRun('Sample');
+        $oShape->getShadow()
+            ->setVisible(true)
+            ->setAlignment(Shadow::SHADOW_BOTTOM_RIGHT)
+            ->setColor(new Color('FF00FF00'))
```
Progi1984: Can you test with a color which is not FF as alpha ?

dkulyk: Added — three colour/alpha pairs now, including `80FF0000` and `66FF0000`.

It caught a real gap: the colour was built here without an alpha, so every shadow read back opaque. It now goes through `loadStyleColor()`, which already reads `a:alpha`.

Progi1984 (approved):

PR #983, opened 2026-09-02
Shape\AbstractGraphic : Set the width and the height a shape is given

### LINKED ISSUE #324: Division by zero by trying to set width and/or height on empty drawing shape
If you try to set width and/or height by empty Drawing Shape instance, the Resize Proportional property is by default true, because of that it triggers „Division by zero“ error. 
`$object = new Drawing();`
`$object->setWidth(100)->setHeight(50); #or $object->setWidthAndHeight(100,50)`
it brings error: "Division by zero"

I fixed this issue and wanted to upload bug-fix (pull-request), but i have no permission or i do something wrong...
In attachment you can find my suggestion to fix this issue
[PHPPresentation.zip](https://github.com/PHPOffice/PHPPresentation/files/787529/PHPPresentation.zip)

dkulyk:
### Description

Fixes #324

`setWidthAndHeight()` divided by the shape's current width and height before checking anything, so calling it on a shape that has neither yet threw a `DivisionByZeroError`:

```php
(new Drawing\File())->setWidthAndHeight(100, 50);   // DivisionByZeroError
```

The two sibling setters were guarded for exactly this in 1.0.0 — `859321fc1` and `f51f8a3fc` — and this one was left as it was. Nothing in `src/` calls it, so it is reachable only from a caller's own code, which is why loading a file stopped throwing while this did not.

The ratios are now worked out only where there is a proportion to keep, meaning all four numbers are there. Where there is not, both dimensions are set as they were asked for.

That second sentence is not a separate change to take or leave: the assignment used to live inside the proportional branch, so guarding the divisions alone would leave the call returning `0x0` on the very shape the issue names, turning the crash into a silent no-op rather than fixing it.

**One case in it is worth a second opinion**, though: a graphic that has dimensions already but does not resize proportionally. That call set neither dimension before and sets both now. `AbstractShape::setWidthAndHeight()`, the method this one overrides, has always assigned both, and the two existing tests for it asserted the width and height the shape already carried from the `setWidth()` and `setHeight()` calls above them, so they passed whether or not the call did anything. If the no-op was deliberate, that branch can keep it.

### Checklist:

> One test for a shape with no dimensions, which errors on `develop`, and one for the non-proportional case, which fails on it.
> `docs/changes/1.3.0.md`.
> Same file.


tests/PhpPresentation/Tests/Shape/AbstractGraphicTest.php
Progi1984: Can you add a test with 0 values ? And rebase ofc 😂

dkulyk:
@Progi1984 Added `testWidthAndHeightWithZero`, six cases with a zero on either side. Before, a shape with a zero dimension threw and a zero asked for was ignored; now both are taken as asked. Rebased onto `master`.

Progi1984 (approved):

PR #988, opened 2026-09-03
Both Readers : Read a line back, and write one that runs right to left the right way round

### LINKED ISSUE #987: PowerPoint2007 and ODPresentation Readers : a line is written by both Writers and read back by neither
Both Writers write a `Line`, and neither Reader reads one. A deck that is loaded and saved again
loses every line it had.

**Measured on `957bb90b8`**, one line and one text shape per deck:

| | in the file | read back |
| --- | --- | --- |
| PowerPoint2007 | `p:cxnSp` with `a:prstGeom prst="line"` | `[RichText]` |
| ODPresentation | `draw:line` | `[RichText]` |

`Reader/PowerPoint2007::loadSlideShapes()` switches on the tag name and has no `p:cxnSp` arm, so a
connection shape falls into the `default` and is dropped. `Reader/ODPresentation::loadSlide()` walks
`draw:frame` only, and a `draw:line` is a shape of the page in its own right rather than the content
of a frame, so it is never reached.

### The Writer mirrors a line that runs right to left

Found while measuring the above, and it has to be fixed for the round trip to close.

An OOXML extent cannot be negative, so a line whose end is behind its start is written from the
other end with `flipH` / `flipV` saying which axis it is mirrored on.
`Writer/PowerPoint2007/AbstractSlide::writeShapeLine()` gets two of the four cases wrong:

| line | branch | written | correct |
| --- | --- | --- | --- |
| down and to the right | `getWidth() >= 0 && getHeight() >= 0` | no flip | no flip |
| up and to the right | `getHeight() < 0` | `flipV` | `flipV` |
| down and to the left | `getWidth() < 0` | **`flipV`** | `flipH` |
| up and to the left | both `< 0` | **nothing** | `flipH` and `flipV` |

So a line drawn right to left arrives mirrored, and one drawn up and to the left arrives mirrored on
both axes -- in PowerPoint today, with no reader involved. `PptSlidesTest::testLine` creates all four
lines but asserts only that some `a:xfrm` carries `flipV`, which the second case satisfies on its
own, so nothing caught it.

The ODPresentation Writer has nothing to get wrong: ODF says a line by the two points it runs
between, so a negative width is simply an `svg:x2` smaller than the `svg:x1`.

PR to follow.

dkulyk:
### Description

Both Writers write a `Line` and neither Reader read one back, so a deck that was loaded and saved again lost every line it had. `Reader/PowerPoint2007::loadSlideShapes()` switches on the tag name and had no `p:cxnSp` arm; `Reader/ODPresentation::loadSlide()` walked `draw:frame` only, and a `draw:line` is a shape of the page rather than the content of a frame.

Measuring the round trip turned up a second defect, in the Writer, which had to go with it. An OOXML extent cannot be negative, so a line whose end is behind its start is written from the other end with `flipH` / `flipV` saying which axis it is mirrored on. `writeShapeLine()` wrote `flipV` for a line that runs right to left, and nothing at all for one that runs up and to the left:

| line | written before | written now |
| --- | --- | --- |
| down and to the right | no flip | no flip |
| up and to the right | `flipV` | `flipV` |
| down and to the left | **`flipV`** | `flipH` |
| up and to the left | **nothing** | `flipH` and `flipV` |

So a line drawn right to left arrived mirrored in PowerPoint, with no reader involved. `PptSlidesTest::testLine` created all four lines but asserted only that some `a:xfrm` carried a `flipV`, which the second case satisfies on its own, so nothing caught it. That test now says which of the three forms it expects, and the two readers each round-trip a line in all four directions.

The ODPresentation Writer needed nothing: ODF says a line by the two points it runs between, so a negative width is simply an `svg:x2` smaller than the `svg:x1`.

Fixes #987

### Checklist:

> `testLineSurvivesTheRoundTrip` in both Reader tests, over the four directions a line can run, and the three flip forms asserted in `PptSlidesTest::testLine`.
> Nothing to describe: `createLineShape()` is documented already and its behaviour is unchanged.
- [x] I have added an entry in the CHANGELOG.md file
> Under Bug fixes.


Progi1984 (commented):
@dkulyk It's ok. Can you rebase it ?

dkulyk:

Progi1984 (approved):

PR #994, opened 2026-09-04
PowerPoint2007 Reader : Reach the package through one accessor

dkulyk:
### Description

Not a functional change: a small refactor, so that the reader stops being written against one
particular archive implementation in nineteen places.

`Reader\PowerPoint2007` named `$this->oZip` nineteen times: once to open the archive, and eighteen
times to read a part out of it from the middle of the parsing code. Opening now happens in
`openPackage()` and reading in `getFromName()`.

| | before | after |
| --- | --- | --- |
| open | `new ZipArchive()` + `open()`, inline in `loadFile()` | `openPackage()` |
| read a part | `$this->oZip->getFromName(...)`, 18 places | `$this->getFromName(...)` |

Nothing changes about what is read, in what order, or what is returned: `getFromName()` hands back
exactly what `ZipArchive::getFromName()` hands back, `false` included. The whole diff is 37 added
lines against 19 removed, in one file.

What it buys is that everything below `loadFile()` is written against "the package this reader is
reading", not against `ZipArchive` -- which is what makes a different container implementation, or a
test double, a matter of those two methods rather than of nineteen call sites.

### Checklist:

> `phpunit`, `phpstan` and `php-cs-fixer` all clean.
> No new behaviour to cover. The refactor is covered by the existing `Reader\PowerPoint2007` suite,
> which reads every sample in `tests/resources/files` through the two new methods and is unchanged.
> Nothing to describe: the two methods are `protected` and no public API moves.
- [ ] I have added an entry in the CHANGELOG.md file
> Deliberately not: nothing observable changes, so there is nothing a reader of the changelog would
> be looking for.

### DIFF
--- a/src/PhpPresentation/Reader/PowerPoint2007.php
+++ b/src/PhpPresentation/Reader/PowerPoint2007.php
@@ -169,6 +169,25 @@ class PowerPoint2007 implements ReaderInterface
     }
 
     /**
+     * Open the package the presentation is read from.
+     */
+    protected function openPackage(string $pFilename): void
+    {
+        $this->oZip = new ZipArchive();
+        $this->oZip->open($pFilename);
+    }
+
+    /**
+     * Contents of one part of the package, or false when it holds no part of that name.
+     *
+     * @return false|string
+     */
+    protected function getFromName(string $path)
+    {
+        return $this->oZip->getFromName($path);
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


Progi1984 (commented):
Could you rebase plz ?

dkulyk:
Rebased on master.

Progi1984 (approved):

PR #996, opened 2026-09-04
PowerPoint2007 Reader : Read a presentation through its main document part

dkulyk:
### Description

Builds on #994, now merged; this is the change itself.

`ppt/presentation.xml` is the name Office happens to write, not one the specification requires. OPC says three things instead: a package points at its main document part through the `officeDocument` relationship, a part says what it is through its content type, and a relationship target is relative to the part that declares it. `Reader\PowerPoint2007` searched for the Office names:

```php
if (is_array($oZip->statName('[Content_Types].xml')) && is_array($oZip->statName('ppt/presentation.xml'))) {
```

and then twenty more times below, `'ppt/' . $pathSlide`, `'ppt/slides/_rels/' . $baseFile . '.rels'`, `'ppt/slides/' . $relImg['Target']`. A package that keeps exactly the same content under other names was refused by `canRead()` and, past that, read as empty.

LibreOffice opens such a file and renders it identically -- same page count, same text, a byte-identical PDF -- which is the check that it is a valid presentation rather than a curiosity.

Three helpers replace the naming:

| helper | what it answers |
| --- | --- |
| `relsOf($part)` | `ppt/slides/slide1.xml` -> `ppt/slides/_rels/slide1.xml.rels` |
| `resolve($relsPart, $target)` | `../media/image1.png` on a slide's rels -> `ppt/media/image1.png` |
| `getFromRel($relsPart, $type)` | the contents of the part a relationship of that type points at |

`canRead()` asks for a main document part whose content type is one of the six a presentation may have (`presentation`, `slideshow` and `template`, each plain and macro-enabled), and `load()` walks outwards from that part. That replaces the twenty-one hardcoded `ppt/...` paths, the four copies of the loop that stripped `..` out of a joined path, and the two `strrpos($target, '../') + 3` hacks; `loadSlide()`, `loadMasterSlide()`, `loadLayoutSlide()` and `loadSlideNote()` carry part names rather than basenames.

Core, custom and thumbnail properties are read through the package-level relationships too, rather than by their `docProps` names.

One bug falls out of it. `presProps.xml` was read and then thrown away:

```php
$pptPresProps = $this->oZip->getFromName('ppt/presProps.xml');
if (false !== $pptPresProps) {
    $this->loadPresentationProperties($pptPresentation);   // the presentation, not the properties
}
```

`loadPresentationProperties()` looks for `/p:presentationPr/p:showPr`, which `presentation.xml` does not hold, so `setLoopContinuouslyUntilEsc()` and the slideshow type never arrived.

### Checklist:

> `phpunit`, `phpstan` and `php-cs-fixer` all clean.
> `Reader\PowerPoint2007\MainDocumentPartTest` reads six samples -- a chart, a group, an image, a slide note, several masters, several slides -- twice: once as they are, and once with the main document part moved to `/deck/main.xml` and its relationships rewritten to point back out of that folder. Both readings must describe the same presentation. On `master` the moved file raises `InvalidFileFormatException`.
> Nothing to describe: the reader accepts strictly more files than before and its API is unchanged.
- [x] I have added an entry in the CHANGELOG.md file
> Under Bug fixes, two entries: the part naming and the presentation properties.


dkulyk:
Draft while it sits on top of #994: the first commit here is that PR. Ready for review as soon as #994 lands, at which point this rebases to a single commit.

src/PhpPresentation/Reader/PowerPoint2007.php
```diff
@@ -187,6 +229,121 @@ protected function getFromName(string $path)
         return $this->oZip->getFromName($path);
     }
 
+    /**
+     * Name of the part holding a part's relationships.
+     *
+     * A part keeps them in a `_rels` folder beside it, in a file named after it, which is where
+     * `ppt/slides/slide1.xml` gets `ppt/slides/_rels/slide1.xml.rels`.
+     */
+    protected static function relsOf(string $part): string
```
Progi1984:
```suggestion
    protected function relsOf(string $part): string
```

src/PhpPresentation/Reader/PowerPoint2007.php
```diff
@@ -187,6 +229,121 @@ protected function getFromName(string $path)
...
+        return ('' === $folder ? '' : $folder . '/') . '_rels/' . basename($part) . '.rels';
+    }
+
+    /**
+     * Resolve a relationship target against the part the relationship was declared on.
+     *
+     * Targets are relative to the folder of their source part, so `../media/image1.png` declared
+     * on `ppt/slides/_rels/slide1.xml.rels` names `ppt/media/image1.png`.
+     */
+    protected static function resolve(string $relsPart, string $target): string
```
Progi1984:
```suggestion
    protected function resolve(string $relsPart, string $target): string
```

src/PhpPresentation/Reader/PowerPoint2007.php
```diff
@@ -187,6 +229,121 @@ protected function getFromName(string $path)
...
+
+        return null;
+    }
+
+    /**
+     * Contents of the part a relationship of the given type points at, or null when there is none.
+     *
+     * @return null|string
+     */
+    protected function getFromRel(string $relsPart, string $type)
```
Progi1984:
```suggestion
    protected function getFromRel(string $relsPart, string $type): ?string 
```

Progi1984 (commented):
No need of static fonction

dkulyk:
@Progi1984 Applied: both are instance methods now, and `getFromRel()` declares `?string`.

Progi1984 (approved):

PR #999, opened 2026-09-07
ODPresentation Writer and Reader : Raise a superscript, and read one back

dkulyk:
### Description

`Font::setBaseline()` reached the `.pptx` writer only, so a superscript saved as `.odp` came back
flat.

| before | after |
| --- | --- |
|  |  |

ODF names the same raise in `style:text-position`, whose first value is a percentage of the font
size where the model holds thousandths of one, so `30000` is written as `30%`. Its second value is
the size the raised character is drawn at, optional, and left out here: `XMLEscapementHeightPropHdl`
in LibreOffice defaults it to `DFLT_ESC_PROP`, 58%, which is what PowerPoint draws.

The attribute goes in through `writeFontStates()`, the one place the underline and the strikethrough
are written, so a chart font carries it too -- as it already does in the PowerPoint2007 Writer.

The reader also answers the two keywords a file written elsewhere may carry. LibreOffice writes
`style:text-position="super 58%"` rather than a percentage, and those two mean "whatever the reader
thinks a superscript is", so they answer with the constants -- the values PowerPoint writes.

Round trip through this branch, and LibreOffice's own `.odp` read by it:

| written | read back |
| --- | --- |
| `30%`, `-25%`, `30.5%` | `30000`, `-25000`, `30500` |
| `super 58%`, `sub 58%` (LibreOffice) | `BASELINE_SUPERSCRIPT`, `BASELINE_SUBSCRIPT` |

Two notes:

- Those constants are ten times too large on `master`; #998 fixes that. This PR does not depend on
  it -- the mapping is the same either way -- but until it lands, `setSuperScript(true)` writes
  `style:text-position="300%"` here for the same reason it writes `baseline="300000"` there.
- The seven **PHP Static Analysis** jobs fail on `master` itself since `phpoffice/common` 1.1.0.
  [PHPOffice/Common#64](https://github.com/PHPOffice/Common/pull/64) dropped the `<\DOMElement>` from
  `XMLReader::getElements()`'s `@return \DOMNodeList`, so a node it yields is now inferred as
  `DOMNode`. 18 errors, all in `Reader/PowerPoint2007.php`, none from this branch:
  `git checkout master && composer update phpoffice/common && vendor/bin/phpstan analyse`
  reproduces them.

Fixes # (issue)

### Checklist:

- [ ] My CI is :green_circle:
> The unit tests pass, 1133 tests and 9227 assertions. Static analysis fails for the reason above, which is `master`'s and not this branch's.
> `ContentTest::testRichTextRunBaseline`, `ODPresentationTest::testFontBaselineSurvivesTheRoundTrip` and `::testFontBaselineFromTextPositionWrittenElsewhere`, and the two chart font-state tests. Every one asserts the document is schema-valid.
> `docs/usage/styles.md`.
> `docs/changes/1.3.0.md`.


src/PhpPresentation/Reader/ODPresentation.php
```diff
@@ -1210,6 +1214,24 @@ protected function loadStylesFile(): void
         }
     }
 
+    /**
+     * The raise `style:text-position` names, in the thousandths of a percent a baseline is held in.
+     * Its `super` and `sub` answer with the values PowerPoint writes.
+     */
+    protected static function baselineFromTextPosition(string $value): int
```
Progi1984:
```suggestion
    protected function baselineFromTextPosition(string $value): int
```

No need to static

src/PhpPresentation/Writer/ODPresentation/AbstractDecoratorWriter.php
```diff
@@ -131,6 +131,19 @@ protected function writeFontStates(XMLWriter $objWriter, Font $font): void
...
+            $objWriter->writeAttribute('style:text-position', self::textPosition($font->getBaseline()));
+        }
+    }
+
+    /**
+     * The raise as a percentage of the font size, where a baseline is held in thousandths of one.
+     * The size it is drawn at is a second, optional percentage, left to the reader -- 58% in
+     * LibreOffice, which is what PowerPoint draws.
+     */
+    protected static function textPosition(int $baseline): string
```
Progi1984:
```suggestion
    protected function textPosition(int $baseline): string
```

Progi1984 (commented):
Some feedback (and rebase ;))

Progi1984 (approved):

